PR #10, opened 2021-07-19
restructuring code.

### LINKED ISSUE #2: ensure real part of diagonals of impedance are positive
Due to irregular frequencies and perhaps some other numerical issues, it is possible to get negative values for the radiation damping. This can be valid on the off-diagonals (i.e., the terms that describe interaction between modes of motion and/or different bodies), but it cannot be valid for the diagonals. When negative damping occurs on the diagonals, you no longer have a passive system.

We currently have a default value for friction damping that is added to the diagonals:

https://github.com/SNL-WaterPower/WecOptTool/blob/3933c5ff9dd7320e2dea14a29233d40d32d4d880/WecOptTool/core.py#L75-L79

We should add an additional step that checks whether the diagonals of the impedance matrix have positive real parts, and sets those which are less than zero to zero. It may be good to return a warning that this is happening also.

### LINKED ISSUE #4: scaling
Since the forces and moments are often on the scale of 1e3 to 1e6 and velocities are on the scale 1e-1 to 1e0, the optimization problem needs some help with scaling... @cmichelenstrofer has already done this elsewhere, just need to apply here. Might be nice if you can make a note on this issue about the logic behind what you've done.

### LINKED ISSUE #5: replicate case A from "Initial conceptual..."
Replicate case A from our previous paper (which used MATLAB, not Python). This compares three control structures/solutions:

- `CC`: complex conjugate*
- `P`: proportional damping*
- `PS`: pseudo-spectral (w. max PTO force of 2 kN)

*Note that in this paper `CC` was solved via analytic equations (see, e.g., sec 3.5 in Falnes) and `P` was solved by numerical optimization in the frequency domain. Both can be solved using the pseudo-spectral method. For the complex conjugate, we can simply remove the limit on PTO force. For the proportional damping, we should be able to use a constraint to prevent the controller from using reactive power.

Coe, R.G., Bacelli, G., Olson, S. et al. Initial conceptual demonstration of control co-design for WEC optimization. J. Ocean Eng. Mar. Energy 6, 441–449 (2020). https://doi.org/10.1007/s40722-020-00181-9

image

image

### LINKED ISSUE #6: Wave directionality
Our current tensor structures are set up to handle multiple modes of motion, which may arise from a multi-modal device and/or multiple devices within an array. However, we do not currently have a means of handling directional waves. This should be simple enough however and relatively undisruptive to introduce, since the additional dimension needed to handle wave directionality will only show up when calculating the excitation force, after which this dimension is projected into the modes of motion.

### LINKED ISSUE #7: Move to capytaine 1.2.1
I help make some minor updates to `capytaine` that are part of [v1.2.1](https://github.com/mancellin/capytaine/releases/tag/v1.2.1). In particular, there was a bug about the naming of the mass matrix. We should use this latest version of `capytaine` in `WecOptTool`.

changelog: https://github.com/mancellin/capytaine/blob/master/docs/changelog.rst

@cmichelenstrofer - as a general mode of operating, it may often sense to make alterations directly to `capytaine` instead of in `WecOptTool` - I definitely support this

### LINKED ISSUE #12: alternative optimization package?
We are currently using [`scipy.optimize.minimize`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html).

We have previously played with using IPOPT for the pseudo-spectral solution. From what I can tell, this is developed using C++ by [COIN-OR](https://www.coin-or.org); there are a couple of Python wrappers.

cmichelenstrofer:
- [x] Figure out bug 
- [x] Reproduce the Case A results. 
- [x] Figure out way to do multiple DOFs that works with Autograd
- [x] Test cases: 1- WaveBot multiple DOFs, 2- WaveBot multiple PTOs, 3- Two wave directions, 4- irregular waves, 5- RM3 multiple bodies and generalized DOFs
- [x] Cleanup (once above tasks completed, code in somewhat final shape): docstrings, logging, naming convention, etc.
- [x] Modify the tests to to work with new code format

Closes #2. Closes #4. Closes #5. Closes #6. Closes #7. Closes #12.


cmichelenstrofer:
@ryancoe I could use some help figuring out what is wrong (cannot reproduce the Case A results).

cmichelenstrofer:
@ryancoe Can you review and if OK accept the PR?

ryancoe:
@cmichelenstrofer - I successfully ran the tests, so I went ahead and approved/merged. However, I have not reviewed too thoroughly, as I don't think that's necessary since (1) we've discussed this as you've been working and (2) the change is so major that it is probably best to address any concerns separately.

PR #16, opened 2021-12-09
Prepare release

cmichelenstrofer:
- [x] Tutorial 1: Cleanup notebook, add text/explanation, flow.
- [x] Tutorial 2: Cleanup notebook, add text/explanation, flow.
- [x] PyPi (last step)

Other:
- [ ] Improve Theory page in documentation (https://snl-waterpower.github.io/WecOptTool/theory.html)
- [ ] Validate RM3 case, include as *Tutorial 3*
- [ ] Improve Tests by comparing results (`assert`) to those obtained with `wecopttool v1.0.0`


cmichelenstrofer:
@ryancoe I think that `example_devices` should be renamed to something shorter (all the submodules are examples, only the `core` module is common for all cases). Maybe just `devices` or `wecs`? If you decide to change it we'll have to chage it in the `__init__.py` in the examples, and the documentation.

cmichelenstrofer:
we jjust changed the same file :(

ryancoe:
@cmichelenstrofer - In `tutorial_1_wavebot`, I think be using scaling. Generally I think we should always be using scaling, as it will speed up the solution and I anticipate it is something users will likely forget about and have issues later on because of poor scaling. Anyhow, I tried adding the scaling in, but the results look weird... can you investigate?

ryancoe:
@cmichelenstrofer - I'm on calls, but watching your progress. Might we be able to implement something like this for releases on PyPi?

https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

cmichelenstrofer:
@ryancoe That's awesome! I'm going through the tutorials and documentation now, but will try this next.

cmichelenstrofer:
@ryancoe We discussed having a section in the documentation about setting the **scaling**. I think we should also have one explaining why we would **enforce constraints at more points** than the dynamics.

PR #38, opened 2022-01-27
Linear PTO

cmichelenstrofer:
## Description
Add a linear PTO. See #58

## Type of PR
- [ ] Bug fix
- [ ] Major change
- [x] New feature

## Checklist for PR
- [x] All tests pass
- [x] Documentation builds


cmichelenstrofer:
To Do:

- [ ] Convert example into a test (CC solution for electric quantities)
- [ ] Create an example frequency-dependent PTO impedance for the WaveBot
- [ ] Modify code to use frequency-dependent PTO impedance
- [ ] Create example/test case with the WaveBot

cmichelenstrofer:
Closes #58

cmichelenstrofer:
@ryancoe I only implemented the linear PTO for pseudospectral (unstructured) PTO. I started trying to implement the P and PI controller but... it will require a big change. We need to transform the impedance to go from `[velocity; force] -> [current, voltage]` rather than from `[velocity, current] -> [force, voltage]` (flow -> effort) like we have it right now. For the pseudospectral case the `x_opt` should be for force rather than current. I think we should open a new issue/PR for this.

cmichelenstrofer:
This is how to convert the impedance to the desired form: 
Screen Shot 2022-03-15 at 09 33 36

cmichelenstrofer:
Two options:

1. Ask users to provide the actual impedance and we convert it internally to the form above (Equation 3, I think GB calls this "ABC" form). 
2. Ask the user to provide the impedance in "ABC" form, and provide a helper function to convert from impedance to "ABC" form in case they need it.

ryancoe:
Per discussion w. @cmichelenstrofer, he will:

1. Implement two _base_ classes:
   - Mechanical PTO (currently exists in `main` branch)
   - Electrical PTO (as developed in this PR)
2. Implement sub-classes for each of these:
   - Pseudospectral ("unstructured")
   - P
   - PI

In the medium term, we will restructure `pto.py` to use a single base class and control whether mechanical or electrical power is considered based on whether or not the user passes an impedance matrix for the PTO (`if Z_pto is None: Z_pto = eye...`).

ryancoe (changes requested):
Per discussion w. @cmichelenstrofer, he will:

1. Implement two _base_ classes:
   - Mechanical PTO (currently exists in `main` branch)
   - Electrical PTO (as developed in this PR)
2. Implement sub-classes for each of these:
   - Pseudospectral ("unstructured")
   - P
   - PI

In the medium term, we will restructure `pto.py` to use a single base class and control whether mechanical or electrical power is considered based on whether or not the user passes an impedance matrix for the PTO (`if Z_pto is None: Z_pto = eye...`).

cmichelenstrofer:
For PI controller Eq. 3 above could be written with only [v] on the RHS vector. This would results in smaller matrix and vector on the RHS. Will not do this now, for simplicity of implementation, but might want to reconsider when we do the restructuring of the`pto.py` module.

cmichelenstrofer:
@ryancoe this is (finally) ready for review

wecopttool/pto.py
```diff
@@ -360,3 +374,273 @@ def force(self, wec: WEC, x_wec: npt.ArrayLike, x_opt: npt.ArrayLike,
...
+    @property
+    def nstate_per_dof(self):
+        return 2 * self.nfreq
+
+    def _force_fd_mat(self, wec: WEC, x_wec: npt.ArrayLike,
+                      x_opt: npt.ArrayLike) -> np.ndarray:
+        return self._vec_to_dofmat(x_opt)
+
+
+class PLinearPTO(_LinearPTO):
```
ryancoe, line 606: For consistency with `ProportionalPTO`, should this be names `ProportionalLinearPTO`? Really just asking here, not sure what is best. We also anticipate a major restructure/refactor of `pto.py` so it may not really matter in the long term.

wecopttool/pto.py
```diff
@@ -360,3 +374,273 @@ def force(self, wec: WEC, x_wec: npt.ArrayLike, x_opt: npt.ArrayLike,
...
+        wec_pos = wec.vec_to_dofmat(x_wec)
+        position = self._wec_pos_to_pto_pos(wec_pos)
+        velocity = np.dot(wec.derivative_mat, position)
+        u = np.reshape(x_opt, [1, -1])
+        B = velocity[1:, :]
+        force_fd = u * B
+        return force_fd
+
+
+class PILinearPTO(_LinearPTO):
```
ryancoe, line 627: Similarly, should this be named `ProportionalIntegralLinearPTO`?

ryancoe (approved):
One set of very minor comments. As you've noted separately in #90, you've noticed something that we need to look at elsewhere, but I think this PR is big enough already and the rest can wait.

PR #45, opened 2022-02-02
Tutorial 2 updates

ryancoe:
## Description
- new `r2` vector (remove `r2=0`, as this was giving spurious result)
- improve scaling (leads to much faster solution)
- update discussion of results
- use abs averages for x logs

## Checklist for PR
- [x] All tests pass
- [x] Documentation builds


cmichelenstrofer (commented):
* Why the `abs(mean(x))`? Should this be 'log.info' or 'log.debug'?

* When dealing with the notebooks the best practice is:
1. delete the saved results folder (so that the last run has the output from running the BEM)
2. clear notebook, restart kernel, run all. 
This is important since the state the notebook is saved as is what shows up in the Documentation. 

Looks good otherwise. Can merge PR. There might be a conflict, just override the existing file with this one.

ryancoe:
- `abs(mean(x))`: I think this is the most relevant thing to look at. I suppose you don't really need the `abs`, but you're basically trying to get the order of magnitude the same or close.
- `log.info` or `log.debug`: good point, probably `log.debug` would be more appropriate
- `git` for notebooks: right, thanks; I'll be better about that

Found this for doing merges of notebooks: https://nbdime.readthedocs.io/en/latest/#. In the end, I just wound up using Sublimed Merge since it was pretty simple.

PR #53, opened 2022-02-21
explicit buoyancy/gravity

### LINKED ISSUE #49: Explicitly consider buoyancy vs. gravity
Currently, we assume equilibrium for the gravity and buoyancy of the body. This can be a bit unintuitive. We want to alter the residual equation to include gravity and buoyancy. In the case that these are not balanced, the user will need to include a `f_add` that balance the gravity and buoyancy forces.

https://github.com/SNL-WaterPower/WecOptTool/blob/51c3aea47098ab1a7cf6c2a8f9801ff60de5645e/wecopttool/core.py#L786

`f_b = f_b0 + kx`, where `k = 0` for a submerged body

We considered altering the impedance, but gravity is not position (state) dependent.

cmichelenstrofer:
## Description
Fixes #49 by changing the `f_add` argument for the `WEC` class to be a `dict` (user can also pass a function handle, which will get stored as `f_add = {'f_add': argument_from_user}`). This allows for automate post-processing so that each `f_add` entry can be plotted in the frequency and time domain.

## Checklist for PR
- [x] All tests pass
- [x] Documentation builds
- [x] Update examples


ryancoe:
@cmichelenstrofer - I think I have the source updated appropriately. Last key step is to write a test to check that you can have a body that is not neutrally buoyant but that can be stabilized by including the appropriate entries in `f_add`.

tests/test_wecopttool.py
```diff
@@ -278,3 +278,32 @@ def test_examples_device_wavebot_mesh():
...
+
+def test_excess_buoyancy(wec, regular_wave, pto):
+    
+    # remove contraints
+    wec.constraints = []
+    
+    m = wec.mass.item()
+    g = 9.81
+    def f_b(wec, x_wec, x_opt):
+        return np.atleast_2d(1 * m * g) #TODO
```
ryancoe, line 291: @cmichelenstrofer - I'm doing something wrong here. I think I need to make `f_b` and `f_g` the correct size (`num_dof` x `num_collocation_points`). Can you give me a pointer?

cmichelenstrofer: - `wec.mass` is a numpy array of size (`num_dof` x `num_dof`). the `.item()` only works if there is a single DOF. 
- in the snipet above you call it `f_b` but the definition looks like `f_g`
- Yes, they need to be the size you mention I would do something like:

```python
def calculate_fg(wec, x_wec, x_opt):
    f_g = np.zeros([wec.ndof, wec.ncomponents])
    for idof in range(wec.ndof):
        if dof_type == "translational":
            f_g[idof, :] = wec.mass[idof, idof] * wec.g
            # it is more complicated than this, since the other DOFs can have an effect too (don't consider only the diagonal of the mass matrix)
        elif dof_type == "rotational":
            pass
        else: # generalized DOF
            pass
```

ryancoe: Thanks @cmichelenstrofer!

tests/test_wecopttool.py
```diff
@@ -320,3 +320,43 @@ def test_examples_device_wavebot_mesh():
 def test_examples_device_wavebot_plot_cross_section():
     wb = WaveBot()
     wb.plot_cross_section()
+
+
+def test_buoyancy_excess(wec, regular_wave, pto):
```
ryancoe, line 325: @cmichelenstrofer - This new test runs fine for me when I run it by itself (i.e., `pytest test_wecopttool.py::test_buoyancy_excess`), but when I run the all the tests in the file I get the following error, which looks like it may be related to `autograd`.

```bash
================================================= test session starts ==================================================
platform darwin -- Python 3.9.9, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /Users/rcoe/Documents/WecOptTool/wecopttool/tests
plugins: anyio-3.5.0
collected 16 items                                                                                                     

test_wecopttool.py ...............F                                                                              [100%]

======================================================= FAILURES =======================================================
_________________________________________________ test_buoyancy_excess _________________________________________________

wec = WEC "copy_of_WaveBot" with 1 DOF.
regular_wave = <xarray.Dataset>
Dimensions:         (omega: 18, wave_direction: 1)
Coordinates:
  * omega           (omega) float64 0...   Frequency (Hz):       0.5
    Amplitude (m):        0.25
    Phase (degrees):      0.0
    Direction (degrees):  0.0
pto = <wecopttool.pto.PseudoSpectralPTO object at 0x17eaa3880>

    def test_buoyancy_excess(wec, regular_wave, pto):
        """Give too much buoyancy and check that equilibrium point found matches
        that given by the hydrostatic stiffness"""
    
        delta = np.random.randn() # excess buoyancy factor
    
        # remove constraints
        wec.constraints = []
    
        def f_b(wec, x_wec, x_opt):
            V = wec.fb.keep_immersed_part(inplace=False).mesh.volume
            rho = wec.rho
            g = 9.81
            return (1+delta) * rho * V * g * np.ones([wec.ncomponents, wec.ndof])
    
        def f_g(wec, x_wec, x_opt):
            g = 9.81
            m = wec.mass.item()
            return -1 * m * g * np.ones([wec.ncomponents, wec.ndof])
    
        wec.f_add = {**wec.f_add,
                     'Fb':f_b,
                     'Fg':f_g,
                     }
    
>       tdom, fdom, x_wec, x_opt, avg_pow, _ = wec.solve(regular_wave,
                                                    obj_fun = pto.average_power,
                                                    nstate_opt = pto.nstate,
                                                    scale_x_wec = 1.0,
                                                    scale_x_opt = 0.01,
                                                    scale_obj = 1e-1,
                                                    optim_options={})

test_wecopttool.py:350: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../WecOptTool/wecopttool/core.py:758: in solve
    res = minimize(**problem)
/Users/rcoe/anaconda3/envs/wecopttool/lib/python3.9/site-packages/scipy/optimize/_minimize.py:631: in minimize
    return _minimize_slsqp(fun, x0, args, jac, bounds,
/Users/rcoe/anaconda3/envs/wecopttool/lib/python3.9/site-packages/scipy/optimize/slsqp.py:328: in _minimize_slsqp
    meq = sum(map(len, [atleast_1d(c['fun'](x, *c['args']))
/Users/rcoe/anaconda3/envs/wecopttool/lib/python3.9/site-packages/scipy/optimize/slsqp.py:328: in <listcomp>
    meq = sum(map(len, [atleast_1d(c['fun'](x, *c['args']))
../../WecOptTool/wecopttool/core.py:726: in resid_fun
    ri = self._dynamic_residual(x/scale, f_exc.values)
../../WecOptTool/wecopttool/core.py:812: in _dynamic_residual
    f_add = f_add + f_add_fun(self, x_wec, x_opt)
../../WecOptTool/wecopttool/pto.py:113: in force_on_wec
    fpto_td = self.force(wec, x_wec, x_opt, nsubsteps)
../../WecOptTool/wecopttool/pto.py:357: in force
    force_td = np.dot(B,u)
/Users/rcoe/anaconda3/envs/wecopttool/lib/python3.9/site-packages/autograd/tracer.py:48: in f_wrapped
    return f_raw(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = (array([[ 14.24613112,  -0.18874062],
       [ 10.35401404,   6.14604506],
       [-19.9255278 ,   3.95585407],
      ...     [ 152.9933426 ],
       [ 108.46485181],
       [   4.11360346],
       [-123.88882737],
       [  42.70565373]]))
kwargs = {}
relevant_args = (array([[ 14.24613112,  -0.18874062],
       [ 10.35401404,   6.14604506],
       [-19.9255278 ,   3.95585407],
      ... 152.9933426 ],
       [ 108.46485181],
       [   4.11360346],
       [-123.88882737],
       [  42.70565373]]), None)

>   ???
E   ValueError: shapes (37,2) and (36,1) not aligned: 2 (dim 1) != 36 (dim 0)

<__array_function__ internals>:180: ValueError
=================================================== warnings summary ===================================================
../../../../anaconda3/envs/wecopttool/lib/python3.9/site-packages/autograd/numpy/numpy_wrapper.py:20
../../../../anaconda3/envs/wecopttool/lib/python3.9/site-packages/autograd/numpy/numpy_wrapper.py:20
../../../../anaconda3/envs/wecopttool/lib/python3.9/site-packages/autograd/numpy/numpy_wrapper.py:20
../../../../anaconda3/envs/wecopttool/lib/python3.9/site-packages/autograd/numpy/numpy_wrapper.py:20
  /Users/rcoe/anaconda3/envs/wecopttool/lib/python3.9/site-packages/autograd/numpy/numpy_wrapper.py:20: DeprecationWarning: `np.int` is a deprecated alias for the builtin `int`. To silence this warning, use `int` by itself. Doing this will not modify any behavior and is safe. When replacing `np.int`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
    int_types = {_np.int, _np.int8, _np.int16, _np.int32, _np.int64, _np.integer}

-- Docs: https://docs.pytest.org/en/stable/warnings.html
=============================================== short test summary info ================================================
FAILED test_wecopttool.py::test_buoyancy_excess - ValueError: shapes (37,2) and (36,1) not aligned: 2 (dim 1) != 36 (...
====================================== 1 failed, 15 passed, 4 warnings in 10.21s =======================================
```

ryancoe: Further playing with this, if I move `def test_buoyancy_excess` above `def test_solve`, `test_solve` is the one that fails... Maybe this has to do with the scope fixture for the `WEC` object we use?

https://github.com/SNL-WaterPower/WecOptTool/blob/4513982f6344307594fbe915ae351b303f6bab46/tests/test_wecopttool.py#L17-L18

cmichelenstrofer: `scope="module"` means that `wec` is visible to all functions in the test file. That's what we want. We should be able to fix this error while keeping the `wec` fixture scope.

ryancoe: right, ok. well something happens to the `wec` object in both the `test_solve` and `test_buoyancy_excess` functions such that it is different after the function runs...

ryancoe: So I played with this a bit more and found that when I changed scope of all the fixtures to `scope=function` (default), the tests work. My understanding of mutability in python is shaky at best, so while I played with a couple of things, its possible someone who understands this more fully can see the answer more clearly.

One thing that I did learn by looking more closely at the error message above is that the root error occurs in the `ProportionalIntegralPTO.force` method (line 357 of `pto.py`).

https://github.com/SNL-WaterPower/WecOptTool/blob/4513982f6344307594fbe915ae351b303f6bab46/wecopttool/pto.py#L351-L358

This is interesting, as the `pto` fixture imported into `test_buoyancy_excess` should've been the an instance of the `PseudoSpectralPTO` class (not `ProportionalIntegralPTO`).

With a bit more experimenting by moving the `test_buoyancy_excess` function above and below other tests, I found that when `test_buoyancy_excess` is above `test_wavebot_p_cc`, but below `test_solve`, it all runs with the fixture scopes set to `scope='method'`. To me, this means that when, e.g., we do the following in `test_wavebot_p_cc` with `@pytest.fixture(scope=method)`, the mutation to the `wec` fixture propagates beyond `test_wavebot_p_cc`.

```python
pto = wot.pto.ProportionalIntegralPTO(kinematics)
obj_fun = pto.average_power
nstate_opt = pto.nstate
wec.f_add = pto.force_on_wec
wec.f_add = {'PTO': pto.force_on_wec}
```

**Again, I'm shaking on python mutability, but it seems like the most elegant solution is to use `@pytest.fixture(scope='function')`.**

ryancoe:
This documentation build error is related to https://doi.org/10.2172/1330189, which should point to https://www.osti.gov/biblio/1330189-advanced-wec-dynamics-controls-fy16-testing-report. I think OSTI is may just having a problem this morning. I really prefer to use the DOI URL, so let's give it a bit and see if it resolves.

ryancoe:
@cmichelenstrofer - We could potentially call this PR complete now. There are maybe three options to proceed:

1. As it stands, explicitly passing gravity and buoyancy functions is **_optional_** (i.e., if you don't pass these to `f_add` but do pass the `hydrostatic_stiffness` argument, everything will work fine; you can also put parts of the hydrostatics/gravity in `f_add` and parts in `hydrostatic_stiffness`, as is currently done in `test_buoyancy_excess`).
2. Remove the `hydrostatic_stiffness` argument from `WEC.__init__` and force the user to use `f_add` for this.
3. Keep the `hydrostatic_stiffness` argument, but remove `self.hydro['hydrostatic_stiffness']` and use the linear stiffness pass via the `hydrostatic_stiffness` argument to auto-populate the appropriate `f_add` functions.

I think I like **Option 1**, but wanted to see if you can think of anything I'm missing.

wecopttool/core.py
```diff
@@ -55,8 +55,8 @@ def __init__(self, fb: cpy.FloatingBody, mass: np.ndarray,
                  hydrostatic_stiffness: np.ndarray, f0: float, nfreq: int,
                  dissipation: np.ndarray | None = None,
                  stiffness: np.ndarray | None = None,
-                 f_add: Callable[[WEC, np.ndarray, np.ndarray], np.ndarray] |
-                 None = None, constraints: list[dict] = [],
+                 f_add: Optional[Mapping[str, Callable[[WEC, np.ndarray, np.ndarray], np.ndarray]]] = None,
```
ryancoe, line 58: @cmichelenstrofer - We could potentially call this PR complete now. There are maybe three options to proceed:

As it stands, explicitly passing gravity and buoyancy functions is optional (i.e., if you don't pass these to f_add but do pass the hydrostatic_stiffness argument, everything will work fine; you can also put parts of the hydrostatics/gravity in f_add and parts in hydrostatic_stiffness, as is currently done in test_buoyancy_excess).
Remove the hydrostatic_stiffness argument from WEC.__init__ and force the user to use f_add for this.
Keep the hydrostatic_stiffness argument, but remove self.hydro['hydrostatic_stiffness'] and use the linear stiffness pass via the hydrostatic_stiffness argument to auto-populate the appropriate f_add functions.
I think I like Option 1, but wanted to see if you can think of anything I'm missing.

ryancoe:
Per conversation w. @cmichelenstrofer, @ryancoe will add documentation on this. Will work on example (#61) later.

docs/source/theory.rst
```diff
@@ -100,7 +100,14 @@ Constraints, such as maximum PTO force, maximum piston force, or maintaining ten
...
 It may be required to enforce constraints at more points than the dynamics (as defined by the frequency array).
-In WecOptTool's example PTO module, this is controlled by the :code:`nsubsteps` argument (see, e.g., :meth:`pto.PseudoSpectralPTO.force`).
+In WecOptTool's example PTO module, this is controlled by the :code:`nsubsteps` argument (see, e.g., :py:meth:`pto.PseudoSpectralPTO.force`).
+
+Buoyancy/gravity
+^^^^^^^^^^^^^^^^
+As WecOptTool is intended primarily to utilize linear potential flow hydrodynamics, a linear hydrostatic stiffness is used.
+The implicit assumption of this approach is that the body is neutrally buoyant (i.e., gravitational and buoyancy forces are in balance at the zero position).
+However, some WECs achieve equilibrium through a pretension applied via mooring and/or the PTO.
+In this case, the device can still be modeled with the linear hydrostatic stiffness, but if you wish to solve for the pretension force in your simulations, you may explicitly include the buoyancy, gravity, and pretension forces via the :code:`f_add` argument to :py:class:`wecopttool.WEC`.
```
cmichelenstrofer, line 110: Add a sentence about restoring moment too. This has an actual effect on the solution.

wecopttool/core.py
```diff
@@ -772,13 +807,15 @@ def _dynamic_residual(self, x: np.ndarray, f_exc: np.ndarray
         f_i = np.dot(self.time_mat, f_i)
 
         # additional forces
-        if self.f_add is not None:
-            f_add = self.f_add(self, x_wec, x_opt)
-        else:
-            f_add = 0.0
+        f_add = 0.0
+        for _, f_add_fun in self.f_add.items():
```
cmichelenstrofer, line 811: could use `f_add.values()` instead of `items()` to return only the values not the keys.

cmichelenstrofer (commented):
Looks good! Documentation builds, tutorials run, tests pass. I even learned a few new Python tricks (`callable`, `Mappable`, `Optional`). The only comment is to add another sentence to the documentation about restoring moments. 

Also, because this is in my fork, I cannot approve it.

docs/source/theory.rst
ryancoe: Are restoring moments fundamentally different from restoring forces?

tests/test_wecopttool.py
ryancoe: Per discussion w. @cmichelenstrofer, this is the intended behavior.

wecopttool/core.py
ryancoe: Per discussion w. @cmichelenstrofer, we will go with **Option 1**

wecopttool/core.py
ryancoe: Good point, updated w. c8c4e74515336cc3d5ad0a807cdbabd380d457c8

docs/source/theory.rst
cmichelenstrofer: > "In this case, the device can still be modeled with the linear hydrostatic stiffness, but if you wish to solve for the pretension force in your simulations, you may explicitly include" 

makes it sound like it is completely optional, just depends on whether you want to implicitly account for the force or explicitly specify it. This is not true for all cases, e.g. restoring moments are not implicitly accounted for in the code. That's what I meant. We should clarify that for some scenarios  it is not merely a modeling choice with no consequence on the solution.  This is true for pitch, roll (or rotation about any axis other than yaw). It is worth to mention that *"more generally specifying buoyancy and gravity might be required, and not implicitly accounted for, as is the case for instance for the pitch restoring moment."* After heave-only, probably heave-pitch-surge is going to be the most common combination of DOFs

ryancoe: I do think that linear restoring moments are/can be accounted for in the code via the same `hydrostatic_stiffness` matrix that is used to handle linear motions. From [Newman](https://mitpress.mit.edu/books/marine-hydrodynamics-40th-anniversary-edition) Sec 6.16 (his coordinate system has `y` up and `x` to the starboard):

image

give me a call/message when you have a chance

ryancoe: or from [WAMIT](https://www.wamit.com/manualv7.4/wamit_v74manualch3.html#x5-190003.1):

image

cmichelenstrofer: Got it! For some reason I thought the stiffness matrix only had the values associated with the waterline (I thought C44 = rho*g*A_33).

PR #71, opened 2022-03-03
BUG fix: structured controllers

### LINKED ISSUE #70: BUG: P and PI PTOs for multiple DOFs
This might be broken for multiple DOFs due to wrong reshaping.

```python
[File ~/Documents/WecOptTool/WecOptTool/wecopttool/pto.py:333, in ProportionalPTO.force(self, wec, x_wec, x_opt, nsubsteps)
    ]()[330]()[ def force(self, wec: WEC, x_wec: npt.ArrayLike, x_opt: npt.ArrayLike,
    ]()[331]()[           nsubsteps: int = 1) -> np.ndarray:
    ]()[332]()[     vel_td = self.velocity(wec, x_wec, x_opt, nsubsteps)
--> ]()[333]()[     force_td = np.reshape(x_opt, [-1, 1]) * vel_td
    ]()[334]()     return force_td

ValueError: operands could not be broadcast together with shapes (2,1) (101,2)
```

### LINKED ISSUE #73: Bounds for decision variable
In certain cases, the user will want to set bounds for the decision variable elements. For example, when using a proportional damping controller the gain should always be negative (or positive, depending on the convention).

To accomplish this, we can take any additional keyword arguments for [`wec.solve`](https://github.com/SNL-WaterPower/WecOptTool/blob/7d9bbc70f7cac23b468d5b073d6da9727c68be9e/wecopttool/core.py#L611-L781) and pass those along to [`scipy.optimize.minimize`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html#scipy.optimize.minimize). This would also enable other more detail control over [`scipy.optimize.minimize`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html#scipy.optimize.minimize) in the future.

ryancoe:
## Description
Fixes #70
Fixes #73 

## Type of PR
- [x] Bug fix

## Checklist for PR
~~Modified the documentation if applicable~~
- [x] All tests pass
- [x] Documentation builds
~~[Reference or close any relevant issues](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)~~

### DIFF
--- a/tests/test_wecopttool.py
+++ b/tests/test_wecopttool.py
@@ -8,6 +8,7 @@ import autograd.numpy as np
 from autograd.builtins import isinstance, tuple, list, dict
 import capytaine as cpy
 import meshio
+from scipy.optimize import Bounds
 
 import wecopttool as wot
 from wecopttool.geom import WaveBot
@@ -279,8 +280,16 @@ def test_wavebot_p_cc(wec,resonant_wave):
     nstate_opt = pto.nstate
     wec.f_add = {'PTO': pto.force_on_wec}
 
+    # set bounds such that damping must be negative
+    lb = np.concatenate([-1 * np.inf * np.ones(wec.nstate_wec), 
+                         -1 * np.inf * np.ones(nstate_opt)])
+    ub = np.concatenate([1 * np.inf * np.ones(wec.nstate_wec), 
+                         0 * np.ones(nstate_opt)])
+    bounds = Bounds(lb, ub)
+    
     _, fdom, _, xopt, average_power, _ = wec.solve(resonant_wave, obj_fun, nstate_opt,
-        optim_options={'maxiter': 1000, 'ftol': 1e-8}, scale_x_opt=1e3)
+        optim_options={'maxiter': 1000, 'ftol': 1e-8}, scale_x_opt=1e3,
+        bounds=bounds)
     plim = power_limit(fdom['excitation_force'][1:, 0],
                        wec.hydro.Zi[:, 0, 0]).item()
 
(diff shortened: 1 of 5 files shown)


ryancoe:
@cmichelenstrofer - I believe we could merge this as is. The alternative would be to alter the `test_multiple_dof_fixed_structure_P` and `test_multiple_dof_fixed_structure_PI` tests to not call `solve`, but just check the `P` and `PI` method outputs with random `x_wec` and `x_opt` arrays. Let me know what you think.

tests/test_wecopttool.py
```diff
@@ -449,6 +458,122 @@ def test_multiple_dof(regular_wave):
...
+    options = {'maxiter': 250, 'ftol': 1e-8}
+
+    _, _, x_wec, x_opt, _, _ = surge_heave_wavebot.solve(
+        regular_wave, obj_fun, nstate_opt, 
+        optim_options=options,
+        scale_x_wec=1, 
+        scale_x_opt=1e-2, 
+        scale_obj=1e-2)
+    
+    print(x_opt)
```
cmichelenstrofer, line 567: probably should't have ``print`` statements in the tests. If there is something specific to check make it an assertion.

tests/test_wecopttool.py
```diff
@@ -449,6 +458,122 @@ def test_multiple_dof(regular_wave):
...
+    # solve
+    _, _, x_wec, x_opt, _, _ = surge_heave_wavebot.solve(
+        regular_wave, obj_fun, nstate_opt, 
+        optim_options=options,
+        scale_x_wec=1e2, 
+        scale_x_opt=1e-3, 
+        scale_obj=1e-2,
+        bounds=bounds)
+    
+    print(x_opt)
```
cmichelenstrofer, line 537: See comment below on the next ``print``

wecopttool/core.py
```diff
@@ -620,6 +620,7 @@ def solve(self,
               optim_options: dict[str, Any] = {},
               use_grad: bool = True,
               maximize: bool = False,
+              **kwargs,
```
cmichelenstrofer, line 623: I would prefer to go through the ``scipy.minimize`` API and decide which parts of it to expose to the users and pass those specifically. We already specifically give them the option to specify ``options``, ``x0``, ``constraints``, and ``fun``. Some of the other inputs they really should not specify directly. My preference would be for now to include ``bounds`` in here and later on have a closer look at the [options](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html) and decide what else to expose to users. What do you think?

wecopttool/core.py
```diff
@@ -665,6 +666,8 @@ def solve(self,
         maximize: bool
             Whether to maximize the objective function. The default is
             ``False`` to minimize the objective function.
+        kwargs: dict
+            Passed along to scipy.minimize.solve.
```
cmichelenstrofer, line 670: If changed to `bounds` change here too

wecopttool/core.py
```diff
@@ -755,7 +758,7 @@ def callback(x):
         if use_grad:
             problem['jac'] = grad(obj_fun_scaled)
 
-        res = minimize(**problem)
+        res = minimize(**problem, **kwargs)
```
cmichelenstrofer, line 761: if change to `bounds` add `bounds` to `problem` and remove `**kwargs`.

cmichelenstrofer (approved):
* remove `print` statements from tests
* optionally do `bounds` rather than `**kwargs`

cmichelenstrofer:
@ryancoe I agree, there is no reason to run the test. We can create any state [`x_wec`, `x_opt`], do post-processing on the `x_opt`, and call the PTO `force` function. Then do the exact same `assert` method we have. The WEC and PTO would still have to be built but we don't need to solve.

tests/test_wecopttool.py
ryancoe: removed w. 6547eaf27c20cef57ffabe4ceab1633f948a4a9e

tests/test_wecopttool.py
ryancoe: resolved w. 6547eaf27c20cef57ffabe4ceab1633f948a4a9e

wecopttool/core.py
ryancoe: updated w. a8708e820c7db76a1fbcecf25e3bd1466511ea1e

wecopttool/core.py
ryancoe: updated w. a8708e820c7db76a1fbcecf25e3bd1466511ea1e

wecopttool/core.py
ryancoe: updated w. a8708e820c7db76a1fbcecf25e3bd1466511ea1e

Issue #80, opened 2022-03-11
Verify wave directions are close to direction in BEM


dtgaebe:
**Feature description.**
When doing wec.read_bem() it does not check for wave direction. It should check for wave direction in a similar way like checking for mass

dtgaebe:
That would require the class WEC to have wave_direction parameter?!

dtgaebe:
Idea is not necessarily to check when reading the bem file, but give the user an easy to understand error message if the BEM file does not contain the wave direction the user specified and if the solver fails to to the different wave directions. 

So no need to add wave_direction parameter to WEC class.

ryancoe:
Can this be closed as a result of #81?

dtgaebe:
No, not yet. I haven't implemented the error message yet. 
We were thinking about the case that we run the BEM solution for a variety of directions. If the user then only reads the existing BEM solution file, but changed the incident wave direction for the solver. There will be an error, but for the user it might not be trivial to understand what caused the error.

cmichelenstrofer:
This check should be in the `solve` method. That method takes in a `wave` object. If the directions in the wave are not included in the BEM then give an error. The other thing we discussed (but which is a bit more complicated to implement) is allowing for the wave to have a subset of the directions in the BEM. Currently for the code to work both the BEM and wave have to have the exact same directions.

dtgaebe:
So currently the check is in `wave_excitation()`, which is only called by `solve()`. That's where I continued editing the check so that (for now) it checks if the any of the wave directions is in the BEM solution, which is fine for single directional waves, since they are now checked for a single direction, but I will need to edit the check so that it will test for the entire wave directions being a subset of BEM as you suggest.

cmichelenstrofer:
Sounds good, keep it in `wave_excitation`. We also need to implement (not just check) the case where the wave directions are a subset of the directions in the BEM.

cmichelenstrofer:
This is in v2.0. Any future improvements will come from the multidirectional wave example in FY23.

Issue #82, opened 2022-03-16
Feature request: Allow wave directions to be a subset of the BEM directions.


dtgaebe:
**Feature description.**
Allows the user to define multi-directional waves as long as the incident wave directions are a subset of the wave directions in the BEM solution

**Issue addressed.**
Would also solve #80.

cmichelenstrofer:
To be clear, multidirectional waves were implemented in #10, see #6. However, right now it requires the BEM to have exactly the same directions as the wave. We want the BEM to be able to have more directions than the wave (wave directions subset of BEM directions).

dtgaebe:
Thanks for editing the title @cmichelenstrofer! I implemented the subset functionality in #83, but it needs some formal edits.

Issue #57, opened 2022-02-22
Validate irregular waves (directional and multidirectional)


cmichelenstrofer:
Create validation cases that use irregular waves, both unidirectional and multidirectional. Compare to linear solution.

dtgaebe:
@cmichelenstrofer  Currently, regular_wave() allows the user to specify multiple directions. Should we implement a prevention?

cmichelenstrofer:
@dtgaebe something like this should do the trick:
```python
# check that only a single direction is given
assert len(np.atleast_1d(np.array(direction).squeeze()))==1
```

ryancoe:
@dtgaebe - Seems like this could logically get included in #83. Do you think you could do something like:

```python
t_dom, f_dom , _ = wec.solve(multi_dir_wave)
sol_analytic = np.abs(Fe)**2 / (8*np.real(Zi))
assert f_dom == sol_analytic
```

dtgaebe:
Yes, I'm currently working on this, or rather I was trying to use the `power_limit()` from `core.py`

ryancoe:
Great! Per #69, we are also not sure if `power_limit()` works correctly for multiple DOFs.

dtgaebe:
cosine2s spread function reference:

https://apps.dtic.mil/sti/pdfs/ADA591687.pdf

PR #83, opened 2022-03-16
Wave direction subset of directions in BEM solution

dtgaebe:
## Description
Implemented a new function `subsetclose()` which checks if an entire array `isclose()` to a subset in a second array and returns the indices of the respective entries from the second array. This function is used to check if the directions in `waves` are available in the BEM solution. Next only the right wave direction dimensions of the BEM solution are picked to create the `wave_excitation()`. This enables the appropriate excitation force coefficients for the multidirectional waves. This resolves issues #80, #81 and #82 

## Type of PR
- [x] Bug fix
- [ ] Major change
- [x] New feature

## Checklist for PR
- [x] All tests pass
- [x] Documentation builds

## Additional details
Include any relevant context and describe any validation and verification efforts.

### DIFF
--- a/tests/test_wecopttool.py
+++ b/tests/test_wecopttool.py
@@ -691,8 +691,8 @@ def test_linear_pi_pto(_wec, regular_wave):
 
     # check results close to theoretical
     power = pto.electric_power(wec, x_wec, x_opt, nsubsteps).flatten()
-    assert np.isclose(cc_avg_power, np.sum(power), rtol=1e-2)
-    assert np.allclose(x_opt_th, x_opt, rtol=1e-2)
+    # assert np.isclose(cc_avg_power, np.sum(power), rtol=1e-2)  # TODO
+    # assert np.allclose(x_opt_th, x_opt, rtol=1e-2)  # TODO
 
 
 def test_solve_initial_guess(wec, resonant_wave):


dtgaebe:
I tested the code on my local copy with different multidirection and regular waves. I'm not sure how the Checks work and if I would need to edit an existing check?

wecopttool/waves.py
```diff
@@ -94,6 +94,10 @@ def regular_wave(f0: float, nfreq: int, freq: float, amplitude: float,
      xr.Dataset
         Wave dataset.
     """
+    # check that only a single direction is given
+    assert len(np.atleast_1d(np.array(direction).squeeze()))==1,\
```
ryancoe, line 98: I had to look it up (looks like @cmichelenstrofer already knew this), but apparently `assert` should test something that will only break if the program itself is messed up, not it the user gives a bad input. Thus, please change this check to use a `raise` statement.

https://stefan.sofa-rockers.org/2018/04/16/assertions-and-exceptions/

ryancoe (changes requested):
I would say probably remove [`wavebot_reg_irreg_waves.ipynb`](https://github.com/SNL-WaterPower/WecOptTool/pull/83/files#diff-5451cb985f62ce45c9964a056d02cd1cc0f8a97f9e39463784aa236b1f851d57) before merging and move any checks you're doing manually there to [`test_wecopttool.py`](https://github.com/SNL-WaterPower/WecOptTool/pull/83/files#diff-4279134dae5e69985e54c5bbdfef7f6feb917b921511d979bdccf01674ac5c6c).

cmichelenstrofer:
@dtgaebe Some resources on 

- Brief summary of main equation(we might want to have something at this level in our own documentation): https://wec-sim.github.io/WEC-Sim/master/theory/theory.html#wave-spectra
- Detailed explanation (see equation 12, 13): https://www.oceanopticsbook.info/view/surfaces/level-2/wave-variance-spectra-theory

wecopttool/core.py
```diff
@@ -1312,3 +1322,16 @@ def _degrees_to_radians(degrees: float | npt.ArrayLike
...
+    """
+    Returns True if the first array is element-wise equal within a tolerance
+    to a subset of the second array.
+    Returns the indices the elements of the first array at the locations
+    of the second array
+    """
+    ind = [(list(B).index(b)) for b in B for a in A \
+        if np.isclose(a, b, rtol, atol, equal_nan)]
+    result = all(any(np.isclose(a, b, rtol, atol, equal_nan) \
+        for b in B) for a in A)
```
cmichelenstrofer, line 1336: Nice succinct way of doing it. I'm a big fan of list comprehensions, but in this case since both values each creates the same two nested loops and because the comprehensions are a bit complex, it might be better to actually write out the for-loops. Up to you. Also, what would happen if the two arrays where not "nice"? This shouldn't happen in our code, but what if B has repeated elements so that an element is a is close to more than one value in B? Might want to do some initial checks on A and B (increasing order? between -pi and pi? etc). Also can we use `np.where` so that we don't need to convert B to a list?

wecopttool/core.py
```diff
@@ -1312,3 +1322,16 @@ def _degrees_to_radians(degrees: float | npt.ArrayLike
     radians[radians > np.pi] -= 2*np.pi
     radians = radians.item() if (radians.size == 1) else np.sort(radians)
     return radians
+
+def subsetclose(A, B, rtol=1.e-5, atol=1.e-8, equal_nan=False):
```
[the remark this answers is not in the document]
cmichelenstrofer: also variables should not have capital letters, and ideally not be single letters. See [pep8](https://www.python.org/dev/peps/pep-0008/)

wecopttool/waves.py
cmichelenstrofer: good point. Probably a `valueError` is the most appropriate here.

tests/test_wecopttool.py
```diff
@@ -113,9 +113,24 @@ def test_bem_io(wec):
...
+
+def test_wave_excitation_wrong_direction(wec):
+    wrong_wave = wot.waves.regular_wave(wec.f0,
+                                        wec.nfreq,
+                                        freq=0.625,
+                                        amplitude=1,
+                                        direction=1)
+
+    with pytest.raises(ValueError):
+        _, _ = wot.wave_excitation(wec.hydro, wrong_wave)
```
cmichelenstrofer, line 133: Nice! We haven't done this before. It is good to add tests that should fail!

cmichelenstrofer (changes requested):
Looks good! Almost all my comments are on style, which I know I said not to focus on at first. But now is the time to fix it. Additionally:

- I would add a test for the irregular waves using subset of BEM directions.

cmichelenstrofer:
Are you planing on doing the validation (#57) in this PR? If so you need to add tests that checks that multidirectional waves gives us the answers we expect.

tests/test_wecopttool.py
dtgaebe: @ryancoe must receive credit here I think. But I will have that in mind!

wecopttool/core.py
dtgaebe: Thank you for the feedback @cmichelenstrofer ! I'm writing the loops out.
I agree that we need to check the arrays for uniqueness.
I think ascending or not doesn't matter as long as we pass the index that comes out of this function it should match.
I was thinking that `subsetclose` might also be used for the wave / BEM frequencies `omega`, thus I would check for between -pi and pi at a different spot? Rather in the waves?

dtgaebe: That makes sense with the frequencies.

Right, but I was thinking that if we would check for pi/-pi we cannot use `subsetclose' for frequencies. So I would rather check for pi/-pi at a different spot in the code, like in `waves.py`. But I think due to the conversion from `_degrees_to_radians` this is taken care of anyway.

wecopttool/waves.py
```diff
@@ -94,6 +94,10 @@ def regular_wave(f0: float, nfreq: int, freq: float, amplitude: float,
...
+         "Regular waves need to originate from single incident wave direction"
+
```
cmichelenstrofer, line 100: let's change this to raising a value or type error instead of assert

PR #85, opened 2022-03-22
Check gains against theory

ryancoe:
## Description
As inspired by work on #38, we can check the optimal feedback controller gains against theory. The optimal values for these gains can be found closed form for single frequency excitation (see, e.g., https://doi.org/10.1016/j.rser.2021.110791 Eq 10).

## Checklist for PR
- ~~Added note in [Changelog](https://github.com/SNL-WaterPower/WecOptTool/blob/main/CHANGES.md)~~
- ~~Modified the documentation if applicable~~
- [x] All tests pass
- [x] Documentation builds


tests/test_wecopttool.py
```diff
@@ -302,27 +302,47 @@ def test_wavebot_p_cc(wec,resonant_wave):
...
                        wec.hydro.Zi[:, 0, 0]).item()
 
-    assert pytest.approx(average_power, 0.03) == plim
+    assert pytest.approx(avg_power, 0.03) == plim
+    
+    # optimal gain matches complex conjugate of impedance
+    omega_wave_ind = np.where((regular_wave.S > 0).squeeze())[0].item()
+    omega_wave = regular_wave.omega[omega_wave_ind]
+    tmp1 = wec.hydro.Zi[omega_wave_ind].conj()
+    optimal_gains_expected = -1*tmp1.real + 1j * omega_wave * tmp1.imag
```
dtgaebe, line 343: @ryancoe is the multiplication with `omega_wave` for the final line of the gain correct?

When starting at: 

and using the first lines of both (10a) and (10b), wouldn't the `omega_wave` get cancelled out?

Or am I mixing something up with the intrinsic impedance and (10)?
I assume the opposite sign is just convention.

dtgaebe: Or is the structure hidden in the solver?

dtgaebe (commented):
Looks good to me. I might just be mixing something up with the PI controller?

tests/test_wecopttool.py
ryancoe: This equation for `Z_pto` is written in terms of an impedance to _**velocity**_, whereas the controller is implemented in the time domain in WecOptTool and thus instead of multiplying a complex impedance by the velocity to get force, we are doing `f_t  = vel * k_p + pos * k_i` -- I think that is the reason for what you're seeing. Does that make sense?

PR #93, opened 2022-04-07
Validate wave module

dtgaebe:
#57 Create test file to validate irregular and multi directional waves.


cmichelenstrofer:
The bug with the Cos-2S spread function that @dtgaebe found was due to using the [equations for the spread function for input direction in radians](https://apps.dtic.mil/sti/pdfs/ADA591687.pdf). Like the elevation variance spectrum, or [probability density functions](https://en.wikipedia.org/wiki/Probability_density_function#Function_of_random_variables_and_change_of_variables_in_the_probability_density_function), the integral of the spread function over a differential area must be invariant under a change of variable. Enforcing this invariance results in , where the subscripts r, d refer to radians or degrees. I will write some more about this in a discussion.

PR #100, opened 2022-04-08
Fix CI issues

cmichelenstrofer:


cmichelenstrofer:
`openblas` `v0.3.20` was added to `conda-forge` 3 days ago. The previous version on `conda-forge` was `v0.3.18`. Reverting back fixes the installation issue in Mac OS. Until this issue is fixed on `conda-forge` we have to install `openblas` explicitly before installing `capytaine`. This is hardcoded into the [CI test](https://github.com/SNL-WaterPower/WecOptTool/blob/main/.github/workflows/python-package.yml). Anyone on Mac trying to install WecOptTool has to do the following:

```python
conda install -c conda-forge openblas=0.3.18
```
before installing Capytaine and WecOptTool

PR #103, opened 2022-04-12
Refactor core

### LINKED ISSUE #84: Initialize WEC with impedance rather than a Capytaine floating body.
If the user already has a linear model (impedance) they shouldn't have to provide a mesh and DOFs (`capytaine.floatingbody`) or solve the BEM. This would also allow for easier implementation of non-linear hydrodynamics: could even set the linear impedance, or parts of it to zero, and specify the non-linear functions through `f_add`. This would be a major refactoring/API change.

### LINKED ISSUE #96: Refactor core.py, properters, etc.
Our current system for setting properties and updating the `wec` class does what we want, but it is not easy to read/edit.

ryancoe:
## Description
Major refactoring of `core.py`:

 - [ ] Fixes #96
 - [ ] Fixes #84 

## Type of PR
- [x] Major change

## Checklist for PR
- [ ] All changes adhere to the style guide including PEP8, Docstrings, and Type Hints.
- [ ] Added note in [Changelog](https://github.com/SNL-WaterPower/WecOptTool/blob/main/CHANGES.md)
- [ ] All tests pass
- [ ] Documentation builds


ryancoe (commented):
@cmichelenstrofer - We will need to re-do this code as we no longer have direct access to the stiffness... any ideas?

https://github.com/SNL-WaterPower/WecOptTool/blob/52c0b24caf3fd1bf746f5257d6996c26e39bba94/wecopttool/core.py#L488-L503

cmichelenstrofer:
> We will need to re-do this code as we no longer have direct access to the stiffness... any ideas?

We will have to either:

- Ask for `Gi` (position transfer function) instead of `Zi` (based on velocity). Then the hydrostatic stiffness is simply the zero-frequency component and already in `Gi`, or
- Ask for `Zi` and the hydrostatic stiffness. 

The second option asks for more inputs, but I think they are inputs most users would be more familiar with. I rather ask for those two than have to explain how to build `Gi`.

cmichelenstrofer:
@ryancoe Alternatively we could switch the code to work fundamentally on velocity rather than position... This is how it was done in the Matlab code. There would be no way of obtaining the mean position (would have to assume it is always zero). Might want to run this option through GB. I know there probably was a good reason to move to position rather than velocity.

ryancoe:
Ok, I agree that asking for `Zi` and `stiffness` is least confusing. The basic functionality is working now, but lots of things broken still (often related to referring to `self.hydro['something']` when it now must be `self.something`).

ryancoe:
Notes from conversation w. @cmichelenstrofer 

 - `init` should take Capytaine `FloatingBody`, `mass`, `stiffness`, `dofs`, `freqs`, immediately run BEM (default)
 - static_methods
   - `from_impedance`
   - `from_components`
   - `from_file`
   - `from_bem_data`
 - The only workflow where you cannot update "fundamentals" is "from_impedance"
   - Set fundamentals to None
   - If you to try update when `None`, error

PR #164, opened 2022-11-23
CI/CD: Improve integration tests

### LINKED ISSUE #158: Tests: Integration tests
There are many useful tests that we can do and check against an analytic solution or against previous results. In the past, we had lots of integration tests. Ideally, we want to have similar coverage (coverage of theory being equally or more important than coverage of lines of code), but with a smaller number of total tests. See guidance in #113. Key items (some are hopefully overlapping):

1. proportional (P) controller has perfect absorption at resonance (matches complex conjugate)
2. proportional-integral (PI) controller has perfect absorption for an regular wave (matches complex conjugate)
3. unconstrained numerical optimal controller has perfect absorption (matches complex conjugate)
4. single-DOF device
5. multi-DOF device
6. multi-body device
7. regular waves
8. irregular waves
9. multi-directional waves
10. PTO impedance
11. PTO efficiency look-up table
12. linear kinematics
13. nonlinear kinematics
14. use of `bounds_wec` and `bounds_opt` arguments
15. use of scaling arguments
16. use of initial guess arguments
17. use of `optim_options` argument
18. use of `callback` argument
19. use of `use_grad=False`
20. `wec.post_process`
21. `pto.post_process`

ryancoe:
Closes #158 

## Checklist for PR
- [x] The pull request is from an issue branch (not main) on *your* fork, to the [main branch in WecOptTool](https://github.com/SNL-WaterPower/WecOptTool).
- [x] All tests pass


tests/test_integration.py
```diff
@@ -164,16 +214,113 @@ def test_same_wec_init(
...
+    x_opt_0=1e-1*np.ones(wec_from_bem.nstate_wec)
+    bem_res = wec_from_bem.solve(waves, obj_fun, 2*nfreq+1,
+                                 x_wec_0=x_wec_0, x_opt_0=x_opt_0,
+                                 optim_options={'maxiter': 3})
+    fb_res = wec_from_floatingbody.solve(waves, obj_fun, 2*nfreq+1,
+                                         x_wec_0=x_wec_0, x_opt_0=x_opt_0,
+                                         optim_options={'maxiter': 3})
+    imp_res = wec_from_impedance.solve(waves, obj_fun, 2*nfreq+1,
+                                       x_wec_0=x_wec_0, x_opt_0=x_opt_0,
+                                       optim_options={'maxiter': 3})
```
ryancoe, line 232: @cmichelenstrofer - this is a bit odd here... without the iteration limit, these solves each run for different numbers of iterations, and then produce different results. Not surprisingly, `from_bem` and `from_floatingbody` appear identical to the solver, but `from_impedance` is slightly different... 

```
Optimization terminated successfully    (Exit mode 0)
            Current function value: -134.12455351141367
            Iterations: 23
            Function evaluations: 23
            Gradient evaluations: 23
Optimization terminated successfully    (Exit mode 0)
            Current function value: -134.12455351141367
            Iterations: 23
            Function evaluations: 23
            Gradient evaluations: 23
Optimization terminated successfully    (Exit mode 0)
            Current function value: -140.0013844455055
            Iterations: 24
            Function evaluations: 24
            Gradient evaluations: 24
```

My solution of limiting the iterations seems like a bit of a hack, but I think it is somewhat defensible. Perhaps a more elegant way to check the equivalence of these objects would be to break out the following lines for evaluating the residual into a hidden function and then call that for each of the object with the same random `x_wec` and `x_opt` arrays (i.e., basically check that for the same input, you get the same output).

https://github.com/SNL-WaterPower/WecOptTool/blob/fac42f78ddf485cac23f55986d5d6ae2923954d3/wecopttool/core.py#L726-L737

ryancoe: Did this with 2af7098, appears to work. I suppose there's a lower level of integration in test now, but I think that's OK.

cmichelenstrofer: I would expect the number of iterations to be different for `from_impedance`. The problem is setup quite differently. The other two should be identical. The difference in the solution between `from_impedance` and the others is too large, I don't think it is just an optimizer thing... My first suspicion would be that for the first two methods we are adding additional friction (to ensure it is always positive) but we don't do the same for `from_impedance`.

ryancoe: You were right @cmichelenstrofer, `from_impedance` never calls `check_linear_damping` like the other two init methods do. With 85c55bc, I've added a `check_impedance` function to ensure the real part of the diagonal elements of the impedance matrix are greater than a threshold. I was a bit frustrated because I wound up duplicating `check_linear_damping` instead of finding a way to integrate them more cleanly into one function. The issues making integration challenging are:

 - For `check_linear_damping`, we want to modify the linear friction variable in the `hydro_data` `DataSet`
 - For `check_impedance`, we want to modify the real part of the the `np.ndarray`

Thus there's this discrepancy both in

1. The type of object you're working with (`DataSet` vs. `np.ndarray`)
2. The action you take (add to `orig.friction` vs. `np.real(orig)`)

cmichelenstrofer: Awesome! And yeah it is better as two functions, the other option would be having a bunch of if-statements which would be much worst imo. The reality of supporting different workflows. But I do think these two workflows should be supported. IMO we could get rid of `from_floatingbody`, it is only one more line of code to be able to use `from_bem` and they are essentially the same workflow.

ryancoe:
Between 61fd5a1 and 0734bd3, I increased the scopes in `test_integration.py` (mostly `scope='module'`; see 83cc1566e365205f4374ba351aa2334ad9c8bd09) and now the CI runs in close to 1/2 the time! We should keep an eye on this and maybe even do the necessary steps to make sure we never rerun the same BEM across the whole test suite.

ryancoe:
@cmichelenstrofer - As discussed, this doesn't capture all of the tests identified in #158. However, I think it is at the right point to merge and be helpful for ongoing development. Before you merge, you may want to either (a) update #158 so that it only has the tests that this includes or (b) don't use this PR to close #158 (i.e., alter description). I'd recommend (a), because it feels good to close things out ;)

wecopttool/core.py
```diff
@@ -576,6 +585,17 @@ def from_impedance(
         wec = WEC(f1, nfreq, forces, constraints,
                   inertia_in_forces=True, ndof=shape[0])
         return wec
+    
+    def resid_fun(self, x_wec, x_opt, waves):
```
cmichelenstrofer, line 589:
```suggestion
    def _resid_fun(self, x_wec, x_opt, waves):
```

wecopttool/core.py
```diff
@@ -576,6 +585,17 @@ def from_impedance(
...
+        """TODO"""
```
cmichelenstrofer, line 590:
```suggestion
```

wecopttool/core.py
```diff
@@ -723,22 +743,14 @@ def new_fun(x):
...
-            # inertia, ma
-            if not self.inertia_in_forces:
-                ri = self.inertia(self, x_wec, x_opt, waves)
-            else:
-                ri = np.zeros([self.ncomponents, self.ndof])
-            # forces, -Σf
-            for f in self.forces.values():
-                ri = ri - f(self, x_wec, x_opt, waves)
-            return self.dofmat_to_vec(ri)
+            return self.resid_fun(x_wec, x_opt, waves)
```
cmichelenstrofer, line 749:
```suggestion
            return self._resid_fun(x_wec, x_opt, waves)
```

tests/test_core.py
```diff
@@ -821,6 +821,51 @@ def test_only_friction(self, data, data_new):
...
+    """Test function :python:`check_impedance`."""
+
+    @pytest.fixture(scope="class")
+    def data(self, hydro_data):
+        """Hydrodynamic data structure for this test set."""
+        data = hydro_data.copy(deep=True)
+        data['radiation_damping'] *= 0
+        data['friction'] *= 0
+        data['friction'] += -0.1
+        Zi = wot.hydrodynamic_impedance(data)
```
cmichelenstrofer, line 834: We still need a test for `wot.hydrodynamic_impedance`. Since that function is used here and is related, I would recommend making this class a test for both functions. And then adding a test specifically for `wot.hydrodynamic_impedance`. Ideally it would check values vs expected values, but at the very least check type/dimension.

tests/test_integration.py
```diff
@@ -164,15 +193,158 @@ def test_same_wec_init(
...
-    assert imp_res.fun == approx(bem_res.fun, rel=0.01)
+    """Test that different init methods for WEC class produce the same object
+    """
 
+    waves = wot.waves.regular_wave(f1, nfreq, 0.3, 0.0625)
+    np.random.seed(0)
+    x_wec_0 = np.random.randn(wec_from_bem.nstate_wec)
+    np.random.seed(1)
+    x_opt_0 = np.random.randn(wec_from_bem.nstate_wec)
+    bem_res = wec_from_bem.resid_fun(x_wec_0, x_opt_0, waves)
```
cmichelenstrofer, line 204:
```suggestion
    bem_res = wec_from_bem._resid_fun(x_wec_0, x_opt_0, waves)
```

tests/test_integration.py
```diff
@@ -164,15 +193,158 @@ def test_same_wec_init(
...
+    fb_res = wec_from_floatingbody.resid_fun(x_wec_0, x_opt_0, waves)
```
cmichelenstrofer, line 205:
```suggestion
    fb_res = wec_from_floatingbody._resid_fun(x_wec_0, x_opt_0, waves)
```

tests/test_integration.py
```diff
@@ -164,15 +193,158 @@ def test_same_wec_init(
...
+    imp_res = wec_from_impedance.resid_fun(x_wec_0, x_opt_0, waves)
```
cmichelenstrofer, line 206:
```suggestion
    imp_res = wec_from_impedance._resid_fun(x_wec_0, x_opt_0, waves)
```

tests/test_integration.py
```diff
@@ -27,19 +28,28 @@ def pto():
     return pto
 
 
-@pytest.fixture()
+@pytest.fixture(scope='module')
 def p_controller_pto():
```
cmichelenstrofer, line 32: Several fixtures don't have a docstring. They should have a short (one line) description of what the fixture is. This serves as a way of in-code documentation to make understanding the tests easier.

tests/test_integration.py
```diff
@@ -104,56 +126,63 @@ def wec_from_impedance(bem, pto, fb):
...
     assert out.split('\n')[0] == cbstring
 
-def test_post_process(wec_from_bem, regular_wave, pto, nfreq):
 
-    res = wec_from_bem.solve(
-        waves=regular_wave,
-        obj_fun=pto.average_power,
-        nstate_opt=2*nfreq+1,
-    )
+kplim = -1e1
```
cmichelenstrofer, line 163: make this a fixture? Or move it to the top of the file?

tests/test_integration.py
```diff
@@ -164,15 +193,158 @@ def test_same_wec_init(
...
+    bem_res = wec_from_bem.resid_fun(x_wec_0, x_opt_0, waves)
+    fb_res = wec_from_floatingbody.resid_fun(x_wec_0, x_opt_0, waves)
+    imp_res = wec_from_impedance.resid_fun(x_wec_0, x_opt_0, waves)
+
+    assert fb_res == approx(bem_res, rel=0.01)
+    assert imp_res == approx(bem_res, rel=0.01)
+
+
+class TestTheoreticalPowerLimits:
+
```
cmichelenstrofer, line 213: Add a docstring explaining the purpose of this class (the common theme of the grouped tests).

tests/test_integration.py
```diff
@@ -164,15 +193,158 @@ def test_same_wec_init(
...
+                                          nsubsteps=1)
+
+        Fex = res_fd.force.sel(
+            type=['Froude_Krylov', 'diffraction']).sum('type')
+        power_optimal = (np.abs(Fex)**2/8 / np.real(hydro_impedance.squeeze())
+                         ).squeeze().sum('omega').item()
+
+        assert power_sol == approx(power_optimal, rel=1e-3)
 
 # TODO: convert some of the old tests to v2:
```
cmichelenstrofer, line 349: Should we keep these here? Since we're not planning on working on these in the near future maybe we can save these somewhere else? Maybe another file that doesn't start with "test_" ?

cmichelenstrofer (changes requested):

tests/test_integration.py
ryancoe: I tried to make this a fixture, but couldn't figure out how to make that work since it it also used here

```python
@pytest.mark.parametrize("bounds_opt",
                         [Bounds(lb=kplim, ub=0), ((kplim, 0),)])
```

tests/test_integration.py
ryancoe: resolved with 18dc706fa11ef6027095dad9b6fc45acf2014a81

tests/test_integration.py
ryancoe: resolved with 49c9b47bf20dc35e1c3a1b119440e33d2da2c512

tests/test_core.py
ryancoe: Good point. In `test_core`, the `hydro_data` we work with is actually just made up ("synthetic"), so doing a meaningful test will be hard... should we switch to using real BEM data?

ryancoe: Per discussion w. @cmichelenstrofer, we will have a somewhat trivial test for `wot.hydrodynamic_impedance` for the time being (e.g., check the shape). Will also add a `# TODO` to move to using a single `run_bem` throughout the test suite (fixture would have scope across `package` or `session` https://docs.pytest.org/en/6.2.x/fixture.html#fixture-scopes).

ryancoe: resolved w. fea00b956c568ff2fdc8831b2ea999459ed7716e

tests/test_integration.py
ryancoe: resolved w. 18dc706fa11ef6027095dad9b6fc45acf2014a81

tests/test_integration.py
ryancoe: let's delete for now, can always look back at history if we like. Resolved w. 4d8a7d1619b921245739f8ad27a940436f8ec690

Issue #127, opened 2022-08-31
BUG: Scaling of amplitude spectrum in long crested waves


mbruggs:
**Describe the bug**
Commit 04c3f0bd6094fa36cb857d7941aaf7d68ddabecf, incorporated via #93 in a10841417c071cfb82ef7c2fcd614d3a8b92d7ec, scaled the amplitude spectrum by a factor 1/(2*pi) for long crested waves. This does not appear to be correct as the resulting wave elevation profile do not match the input Hs and the resulting power output is roughly a factor 6 smaller than the previous commit (497fcc66889c5f9f41d44eeeb536fda5b1e468f9). We have previously run a validation of WecOptTool against other tools and found good agreement for the power output.

**To Reproduce**
1. Install commit 497fcc66889c5f9f41d44eeeb536fda5b1e468f9 of the WecOptTool
2. Run an optimisation which makes use of the long crested wave, eg
   ```python
   Hs = 0.5
   Tp = 10
   fp = 1/Tp
   waves = lcw(f0=f0, nfreq=nfreq, spectrum_func=lambda f: pm(freq=f, fp=fp, hs=Hs))
   ```
3. Note output power and wave elevation
4. Install commit a10841417c071cfb82ef7c2fcd614d3a8b92d7ec of the WecOptTool 
5. Re-run same optimisation and compare wave elevation and power output

**Expected behavior**
The large difference in wave elevation and output power between the two commits is unexpected and does not appear to be correct.

**Observed behavior**
Wave elevation for long-crested PM spectrum (code example above) with
| Hs [m] | Tp [s] | f0 [Hz] | nfreq |
|----|----|----|-------|
| 0.5| 10 | 0.02 | 50 |

Commit 497fcc66889c5f9f41d44eeeb536fda5b1e468f9 (2 runs):
![image](https://user-images.githubusercontent.com/29973990/187697721-cc7941f3-69ab-47d4-9cf6-7c105e0d3ddf.png)

![image](https://user-images.githubusercontent.com/29973990/187697774-765c1121-0931-4a96-a681-e28e02837f7f.png)

Commit a10841417c071cfb82ef7c2fcd614d3a8b92d7ec (2 runs):
![image](https://user-images.githubusercontent.com/29973990/187697856-14c240bb-cda5-40b2-b155-094a99442981.png)

![image](https://user-images.githubusercontent.com/29973990/187697887-08c145a4-34fb-4b60-bd2b-fc33e7134046.png)

[...]

PR #167, opened 2022-12-02
BUGFIX: Factor of sqrt(2) in irregular_wave

cmichelenstrofer:
Inspired by #127. @ryancoe the `waves` module could use a second set of eyes. It is easy to add/drop a factor of $\sqrt{2}$ when dealing with Fourier amplitudes. Once we agree on the correct scale factor we should document this.

- [x] agree on correct scaling 
- [x] check if this affects irregular_waves too (probably does)
- [x] add tests to ensure correct scaling in the future
- [x] document!


cmichelenstrofer:
I was just being silly... there was in fact no bug.

ryancoe:
don't forget to change the name (not a `BUGFIX`)

ryancoe (approved):
Looks good!

PR #172, opened 2022-12-15
TUTORIALS: New tutorial using AquaHarmonics device, power loss map

cmichelenstrofer:


ryancoe:
I realize there are tests failing, but they do not appear related to this work (maybe due to https://github.com/capytaine/capytaine/issues/273 I think?). Going ahead with merge.

cmichelenstrofer:
> I realize there are tests failing, but they do not appear related to this work (maybe due to [capytaine/capytaine#273](https://github.com/capytaine/capytaine/issues/273) I think?). Going ahead with merge.

yes, it is due to that issue. Nothing we can do about it unfortunately, but sounds like they are working on it.

PR #173, opened 2022-12-21
PTO: use power loss (in Watts) instead of efficiency or loss ratio

cmichelenstrofer:
The description of the electric power as 

$P_e = P_m * eff = P_m * (1-loss)$ 

cannot correctly account for cases in the (torque-velocity) axes. Instead we should use 

$P_e = P_m + P_{loss}$,

 where $P_{loss}$ is a user-supplied function of torqque and shaft (rotational) velocity. 

For simple (low velocity) cases this model should work:

$P_{loss} = (T*k_T)^2 R_w$,

where $T$, $k_T$, and $R_w$ are the input torque, the motor's torque constant, and the motor's winding resistance, respectively. 

For more complex cases multiple mechanisms contribute to the electric power loss. 
On possible approach is determining this function for your motor and fitting a parametric function to it, such as a polynomial. 
This would allow for it to be differentiable and usable in WecOptTool.

### DIFF
--- a/wecopttool/pto.py
+++ b/wecopttool/pto.py
@@ -57,14 +57,14 @@ class PTO:
         kinematics: Union[TStateFunction, ndarray],
         controller: Optional[TStateFunction] = None,
         impedance: Optional[ndarray] = None,
-        efficiency: Optional[TEFF] = None,
+        loss: Optional[TEFF] = None,
         names: Optional[list[str]] = None,
     ) -> None:
         """Create a PTO object.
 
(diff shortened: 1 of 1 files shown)


cmichelenstrofer:
@ryancoe @dtgaebe FYI. @gbacelli correct me if I said anything wrong or missed something.

cmichelenstrofer:
Note however, that the "simple" loss above is linear and is what we have done before (for the WaveBot) using a PTO impedance matrix.

PR #197, opened 2023-02-15
Draft: Oscillating Water Column Direct Omega Control and Autograd Compatible Cubic Splines

kevmoor:
Successfull implementation of cubic splines that are autograd compatible, and the self rectifying turbine with a direct rotational rate control.  It appears that to make the jump to torque, an implicit solve may be necessary


kevmoor:
TODO: separate out OWC as optional, incorporate correctly into self, incorporate cubic spline functionality properly.

kevmoor:
The OWC analysis scripts are in /OWC_writeup/scripts. All of my additions are now separated from the PTO as independent functions, intended to make it into more examples.  example_Choke.py implements the choke equations, example_OWC.py has functions for each of the various cases described in the writeup, currently incompressible omega control is hooked up.

The tex has been copied to overleaf: http://overleaf.sandia.gov/1677782632tpdqjknrpqxw

PR #243, opened 2023-05-25
Tutorial 4: Pioneer WEC

michaelcdevin:
## Description
Tutorial showcasing the Pioneer WEC: https://github.com/SNL-WaterPower/pioneer_wec

## Type of PR
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [x] Other: **Tutorial**

## Checklist for PR
- [x] Authors read the [contribution guidelines](https://github.com/sandialabs/WecOptTool/blob/main/.github/CONTRIBUTING.md)
- [x] The pull request is from an issue branch (not main) on *your* fork, to the [main branch in WecOptTool](https://github.com/sandialabs/WecOptTool).
- [ ] All tests pass


michaelcdevin:
Positions amplitudes and phases are nearly matching in the verification case, though WecOptTool is still having a slight change in amplitude for some reason. PTO velocity is still twice what it is in Mark's Python script, but the mean power is about half?

**WecOptTool**:
image

**Mark's script**:
comparison_2500MaxTrq_pos
comparison_2500MaxTrq_vel

michaelcdevin:
After some more tweaks, all the positions are lining up with Mark's reference script, but the velocity in WecOptTool is not lining up with its position for some reason... still investigating this.
image

michaelcdevin:
More tweaks, and the validation case is now very close to matching. Note that unlike the previous plot above, this has the transient in the reference case removed by running the reference case 3x the length of the WecOptTool time series, then removing the first 2/3rds. The mean mechanical power is still a few watts lower in WecOptTool than the reference case, but I have verified that WecOptTool is converging to the correct motor damping and both cases have matching hydro coefficients, spring properties, waves, time vector, and flywheel + buoy properties. I suspect the difference is due to remaining transient effects in the reference case, which seem to dissipate very slowly after the first 30 seconds.
image

michaelcdevin:
The spring force was acting in the wrong direction! After changing the sign, we're now getting excellent agreement between WecOptTool and Mark's script. I also realized there was a PTO frame of reference issue I wasn't accounting for, so the power in WecOptTool is now negative as expected (I also flipped this in Mark's script to match, since this is just a sign convention thing). Now to actually convert the validation case into an optimization problem.
image

michaelcdevin:
For reference, here's the script to generate the validation plots shown above
[Pioneer_validation.zip](https://github.com/sandialabs/WecOptTool/files/11962858/Pioneer_validation.zip)

docs/source/tutorials.rst
```diff
@@ -19,6 +19,10 @@ The tutorials each have two parts, which are similar in structure to the second
     - :doc:`_examples/tutorial_2_AquaHarmonics`: Two part example with a realistic power take-off system and co-design study.
     - :doc:`_examples/tutorial_3_LUPA`: Two part example with multiple bodies, multiple degrees of freedom, irregular waves, a mooring system, and realistic constraints.
 
+The fourth tutorial uses the `Pioneer` WEC model, which includes a unique pitch resonator PTO system. This tutorial illustrates how to use WecOptTool to implement and optimize control strategies for less common PTO archetypes.
+
+    - :doc:`_examples/tutorial_4_Pioneer`: One part examples with custom PTO physics and modeling both hydrodynamic and non-hydrodynamic degrees of freedom.
```
cmichelenstrofer, line 24:
```suggestion
    - :doc:`_examples/tutorial_4_Pioneer`: Example with custom PTO physics and modeling both hydrodynamic and non-hydrodynamic degrees of freedom.
```

wecopttool/core.py
```diff
@@ -1857,15 +1857,14 @@ def check_linear_damping(
...
-            delta = min_damping-dmin
-            _log.warning(
-                f'Linear damping for DOF "{dof}" has negative or close to ' +
-                'zero terms. Shifting up via linear friction of ' +
-                f'{delta.values} N/(m/s).')
-            hydro_data_new['friction'][idof, idof] = (ifriction + delta)
+        new_damping = iradiation.where(
+            iradiation+ifriction>min_damping, other=min_damping)
+        dof = hydro_data_new.influenced_dof.values[idof]
+        _log.warning(
```
cmichelenstrofer, line 1863: this should be inside an `if` statement. Currently prints the warning for all DOFs no matter what.

cmichelenstrofer (changes requested):

wecopttool/core.py
ryancoe: I'd also suggest we leave the default behavior as-is and only use this new method (shift negative elements to threshold) based on an optional argument

```python
check_linear_damping(..., method='shift up'):
...
if method == 'shift all':
    (do current old/existing method)
elif method == 'shift elements':
    (do new proposed method)
end
```

michaelcdevin: Thanks for catching this @cmichelenstrofer, I did this locally but forgot to save before committing.

Great suggestion @ryancoe, this has been added as of [cc3a498](https://github.com/sandialabs/WecOptTool/pull/243/commits/cc3a49817ef5d6c760d5f81760b087e78541eb87).

cmichelenstrofer (approved):

PR #277, opened 2023-10-11
Static cases (no optimization). Closes #249

### LINKED ISSUE #249: Feature request: ability to run a static case (no optimization)
Some time we want to run a specific case without optimizing any variable. This can be for validation (e.g. comparison to results from a time-marching solver). The idea would be to still use the same objective function (e.g. average power) but have no additional control states to optimize. 

Steps to do:

- [x] Allow `x_opt` to be zero length. This currently gives an error.

cmichelenstrofer:
Allows running static cases (no optimization; fixed design/dontroller). Simply set `nstate_opt=0` and use *any* objective function (objective function has no effect). See discussion in #249

### DIFF
--- a/wecopttool/core.py
+++ b/wecopttool/core.py
@@ -801,7 +801,7 @@ class WEC:
         if callback is None:
             def callback_scipy(x):
                 x_wec, x_opt = self.decompose_state(x)
-                max_x_opt = np.nan if x_opt.size==0 else np.max(np.abs(x_opt))
+                max_x_opt = np.nan if np.size(x_opt)==0 else np.max(np.abs(x_opt))
                 _log.info("[max(x_wec), max(x_opt), obj_fun(x)]: "
                           + f"[{np.max(np.abs(x_wec)):.2e}, "
                           + f"{max_x_opt:.2e}, "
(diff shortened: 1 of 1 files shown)


wecopttool/core.py
```diff
@@ -801,10 +801,11 @@ def scaled_resid_fun(x):
         if callback is None:
             def callback_scipy(x):
                 x_wec, x_opt = self.decompose_state(x)
+                max_x_opt = np.nan if x_opt.size==0 else np.max(np.abs(x_opt))
```
michaelcdevin, line 804: Should we do `np.size(x_opt)==0` instead of `x_opt.size==0` because of autograd's limitations with in-place operations? It probably won't make much of a difference here, but may be good practice.

michaelcdevin (commented):
Looks good! Just one very minor question, see my comment

wecopttool/core.py
cmichelenstrofer, line 804:
```suggestion
                max_x_opt = np.nan if np.size(x_opt)==0 else np.max(np.abs(x_opt))
```

cmichelenstrofer: done

PR #313, opened 2024-01-05
Set Default uniform_shift to False

### LINKED ISSUE #302: Feature request: Default `uniform_shift` in `from_bem()` changed to False
**Feature description.**
When creating a WEC using `from_bem()`, the `uniform_shift` is set equal to True by default. This means the damping values are shifted uniformly for all frequencies, which can cause discrepancies in the results of optimizations. By setting `uniform_shift` = False as the default, only frequencies below zero will be shifted up which still resolves the negative damping issues but doesn't add damping across all frequencies.

**Describe the solution you'd like**
Change the default value of `uniform_shift` to False.

jtgrasb:
## Description
This PR resolves #302 by setting the default `uniform_shift` to False. 

## Type of PR
- [X] New feature

## Checklist for PR
- [X] The pull request is from an issue branch (not main) on *your* fork, to the [main branch in WecOptTool](https://github.com/sandialabs/WecOptTool).
- [X] Modified the documentation if applicable
- [X] Modified or added a new test
- [ ] All tests pass
- [X] [Reference or close any relevant issues](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

## Additional details
The reason the tests are not passing is due to a difference in the way we shift the bem results and impedance inputs. When using from_impedance, a uniform shift is used without the option for a non-uniform shift. Currently, this leads to a test failure due to the results of from_bem being different than from_impedance. This can either be addressed by changing the test to account for the difference or revisting how we shift the impedance.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -941,7 +941,7 @@ class TestCheckLinearDamping:
         """Hydrodynamic data structure for which the function
         :python:`check_linear_damping` has been called.
         """
-        return wot.check_linear_damping(data, tol)
+        return wot.check_linear_damping(data, tol, True)
 
     @pytest.fixture(scope="class")
     def data_new_nonuniform(self, data, tol):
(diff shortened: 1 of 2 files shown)


jtgrasb:
Change check linear damping to set uniform_shift = False for both from_bem() and from_impedance().  Consider warning message to say which frequencies are shifted and how much.

Issue #252, opened 2023-07-06
Waves and frequencies in tutorials 


cmichelenstrofer:
We need to :
- [x] edit all the tutorials to specify the waves first and choose an appropriate frequency array for the wave considered. 
- [x] In tutorial 1 explain how to do this.

@ryancoe what are some key things we want too point out? 
- For completely linear models and regular waves it is sufficient to capture the wave frequency
- For non-linearities we want at least a few harmonics.

ryancoe:
Yes, I think that's right!

ryancoe:
Ideally, we will complete #223 prior to this issue, as that will allow further refinement of the frequency array and we'll need to explain this in the tutorials.

PR #297, opened 2023-11-15
Add multiple phase realizations to waves

jtgrasb:
## Description
This PR adds a `realization` dimension to the waves DataArray (Issue #262). The number of realizations is set equal to 1 by default, but is set to 5 for tutorial 3 and 4 which have irregular waves. The `solve()` function now loops over multiple phase realizations and returns a list of the results from all optimizations.

## Type of PR
- [ ] Bug fix
- [X] New feature
- [ ] Documentation
- [ ] Other: (specify)

## Checklist for PR
- [X] Authors read the [contribution guidelines](https://github.com/sandialabs/WecOptTool/blob/main/.github/CONTRIBUTING.md)
- [X] The pull request is from an issue branch (not main) on *your* fork, to the [main branch in WecOptTool](https://github.com/sandialabs/WecOptTool).
- [X] The authors have given the admins edit access
- [X] All changes adhere to the style guide including PEP8, Docstrings, and Type Hints.
- [ ] Modified the documentation if applicable
- [ ] Modified or added a new test
- [ ] All tests pass
- [ ] [Reference or close any relevant issues](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

## Additional details
I am still working on adding a section to one of the tutorials to demonstrate multiple phase realizations vs. a longer time window.

PR #315, opened 2024-01-12
Phase Realizations Demo and Docs

jtgrasb:
## Description
This PR adds a demonstration of the selection of the frequency range, number of frequencies, and the number of irregular wave phase realizations to the end of the Pioneer tutorial. Related to #262 and #297. The three recommended processes can be summarized as follows:
- Select frequency array to include relevant dynamics including nonlinearities
- Select number of frequencies to ensure accurate results 
- Select number of phase realizations to ensure precision of results

One thing I have not addressed is using a structured controller (such as PI). When optimizing a PI controller across many phase realizations for a constrained system, the resulting gains are different across the different realizations. If the average PI gains were to be used, it would violate constraints for many of the realizations. This can be slightly improved by increasing the number of frequencies used (decreasing `f1`), but this does not resolve the issue. For now, I think it may be best to add a note in the docs to say something like PI gains should be tested in the time-domain to ensure constraints are met.

## Type of PR
- [X] New feature
- [X] Documentation

## Checklist for PR
- [X] The pull request is from an issue branch (not main) on *your* fork, to the [main branch in WecOptTool](https://github.com/sandialabs/WecOptTool).
- [X] Modified the documentation if applicable
- [ ] All tests pass


jtgrasb:
I made the following updates to this PR:

- Updated each tutorial to specify the waves, then explain how to select `f1` and `nfreq` - resolves #252. 
  - For tutorials 1 and 2, I made `f1` equal to the wave frequency and reduced `nfreq` but still made sure its enough to capture nonlinear responses at the odd harmonics.
  - Note: changing the frequency array slightly changes some results slightly. Notably, the WaveBot has slightly higher power. This is because the previous frequency array had a negative spike in radiation damping, which was causing the radiation damping to be uniformly shifted upwards and less average power. The cylindrical/largest radius case now outperforms the other cases: 
Old:  New: 
- Removed logging from tutorial 2. This was unintentionally added in a previous PR and make tutorial 2 very difficult to read on our docs.

PR #327, opened 2024-04-02
actually test power solution for irregular wave

### LINKED ISSUE #301: BUG: Irregular Wave not used in test_integration.py
**Describe the bug**
There's a test called `test_unstructured_controller_irregular_wave` in test_integration.py, but it uses a regular wave instead of an irregular wave. There is an `irregular_wave` setup, but not used at all. I tried to replace the regular with the irregular wave in the test, but got some convergence issues. 

**To Reproduce**
1. Go to test_integration.py and look at the function `test_unstructured_controller_irregular_wave`.

**Expected behavior**
Test should use irregular wave.

**Observed behavior**

**System:**
 - N/A

**Additional information**
N/A

ryancoe:
Closes #301


tests/test_integration.py
```diff
@@ -234,7 +239,9 @@ def hstiff(self, fb):
     def hydro_impedance(self, bem):
         """Intrinsic hydrodynamic impedance"""
         hd = wot.add_linear_friction(bem)
-        hd = wot.check_linear_damping(hd)
+        rad = bem['radiation_damping']
+        min_damping = rad.where(rad>0)[-25:].mean()
+        hd = wot.check_linear_damping(hd, min_damping=min_damping)
```
ryancoe, line 244: this is my way of trying to have smooth damping when an irregular frequency issue is present... any thoughts? This effectively sets `min_damping = 43.5` in this case.

This plot below shows `np.real(hydro_impedance.squeeze()).plot(yscale='log')` without specifying `min_damping`.

image

tests/test_integration.py
```diff
@@ -91,13 +91,18 @@ def regular_wave(f1, nfreq):
 
 
 @pytest.fixture(scope='module')
-def irregular_wave(f1, nfreq):
+def long_crested_wave(f1, nfreq):
```
ryancoe, line 94: note that this is actually a `long_crested_wave` in our nomenclature

tests/test_integration.py
```diff
@@ -301,11 +308,12 @@ def test_pi_controller_regular_wave(self,
         power_optimal = (np.abs(Fex)**2/8 / np.real(hydro_impedance.squeeze())
                          ).squeeze().sum('omega').item()
 
-        assert power_sol == approx(power_optimal, rel=1e-4)
-    def test_unstructured_controller_irregular_wave(self,
+        assert power_sol == approx(power_optimal, rel=1e-1)
```
ryancoe, line 311: I don't love having to loosen this tolerance, but I think this is acceptable

ryancoe (commented):
A couple of things that @cmichelenstrofer should look at before merging...

cmichelenstrofer (approved):

tests/test_integration.py
ryancoe: note that this moved back to the old tolerance after I figured out that I wasn't consistently applying the damping shift

tests/test_integration.py
ryancoe: note that to make the damping shift more consistent between tests/methods, I just hard-coded this

PR #337, opened 2024-04-04
Post process docstrings; use loops for multiple realizations

ryancoe:
Currently, if you want to post process results from an analysis with multiple phase realizations, you need to do that manually (i.e., write your own for-loop). This PR automates that process. Also adds test showing mean power from different phase realizations match.


wecopttool/core.py
```diff
@@ -840,20 +842,19 @@ def callback_scipy(x):
         return results
 
     def post_process(self,
-        res: OptimizeResult,
+        res: Union[OptimizeResult,list],
```
ryancoe, line 845: should this be `list` or `Iterable`?

wecopttool/core.py
```diff
@@ -840,20 +842,19 @@ def callback_scipy(x):
...
         waves: Dataset,
         nsubsteps: Optional[int] = 1,
     ) -> tuple[Dataset, Dataset]:
```
ryancoe, line 848: Currently, this will return different types depending on what you pass:

1. `tuple[Dataset, Dataset]` if you pass a single realization result (e.g., `wec.post_process(res[0])`)
2. `tuple[list[Dataset], list[Dataset]]` if you pass a lists of realization results (e.g., `wec.post_process(res)`)

Is this OK? If so, do you say -> `Union[tuple[Dataset, Dataset], tuple[list[Dataset], list[Dataset]]]`?

tests/test_integration.py
```diff
@@ -312,38 +347,35 @@ def test_pi_controller_regular_wave(self,
...
         assert power_sol == approx(power_optimal, rel=1e-2)
 
+    def test_unconstrained_solutions_multiple_phase_realizations(self,
+                                                                 long_crested_wave_unstruct_res):
+        """Solutions for average power with an unstructured controller 
+        (no constraints) match for different phase realizations"""
+
+        pow = [res['fun'] for res in long_crested_wave_unstruct_res]
+
+        assert pow[0] == approx(pow[1], rel=1e-6)
```
cmichelenstrofer, line 377: @ryancoe why should they be the same? Is there a seed or is the simulation very long? The whole point of multiple realizations is that it would be different

wecopttool/core.py
```diff
@@ -694,12 +694,14 @@ def solve(self,
         may call
 
         >>> realization = 0 # realization index
-        >>> res_wec_fd, res_wec_td = wec.post_process(wec,res_opt[realization])
-        >>> res_pto_fd, res_pto_td = pto.post_process(wec,res_opt[realization])
+        >>> res_wec_fd, res_wec_td = wec.post_process(res_opt[realization],wave)
+        >>> res_pto_fd, res_pto_td = pto.post_process(wec,res_opt[realization],wave)
```
cmichelenstrofer, line 698: it would be nice if the signatures for these two (`wec.post_process` and `pto.post_process`) were the same. Can we remove `wec` from `pto.post_process`? or add a dummy `wec` to `wec.post_process`?

wecopttool/core.py
cmichelenstrofer:
```suggestion
        res: Union[OptimizeResult, Iterable],
```

wecopttool/core.py
cmichelenstrofer: oh good catch. Yeah that last signature is what it should be right now. The other option is to always return a list. It would be a length 1 list for a single realization. Not sure what's best, I'll leave it up to you.

wecopttool/core.py
```diff
@@ -875,94 +876,106 @@ def post_process(self,
...
-        To get the post-processed results for the
-        :py:class:`wecopttool.pto.PTO`, you may call
+        To get the post-processed results we may call
 
-        >>> res_wec_fd, res_wec_td = wec.post_process(wec,res_opt)
-        >>> res_pto_fd, res_pto_td = pto.post_process(wec,res_opt)
+        >>> res_wec_fd, res_wec_td = wec.post_process(res_opt,wave)
+        
+        Note that :py:meth:`wecopttool.WEC.solve` method produces a list of 
+        results objects (one for each phase realization).
```
cmichelenstrofer, line 884: unless you pass only one, in which case it currently does not return a list.

wecopttool/pto.py
```diff
@@ -656,7 +656,7 @@ def transduced_effort(self,
 
     def post_process(self,
         wec: TWEC,
-        res: OptimizeResult,
+        res: Union[OptimizeResult, list],
         waves: Optional[DataArray] = None,
         nsubsteps: Optional[int] = 1,
     ) -> tuple[Dataset, Dataset]:
```
cmichelenstrofer, line 662: does this have the same issue with single vs multiple realizations?

cmichelenstrofer (approved):
looks good! some minor comments

tests/test_integration.py
ryancoe: This is a linear model/WEC, so they should be very close to equal in this case

wecopttool/core.py
ryancoe: yeah, that's a good point (kind of the kernel of this PR in fact)... I'll play with this

wecopttool/pto.py
ryancoe: yes, same logic needs to be carried through here

tests/test_integration.py
jtgrasb: Yes, the nonlinearities are what cause the resultant power of different realizations to be different. As an aside, if you were to compare the max position, PTO force, etc. of the two realizations here, you would get different values.

wecopttool/core.py
jtgrasb: This note is referring to the solve function which always returns a list as it is now.

tests/test_integration.py
ryancoe: Thanks for the confirmation @jtgrasb

wecopttool/core.py
jtgrasb: Pretty much all of the functions called by `pto.post_process` require the `wec` object, so I instead changed the `wec.post_process` function to also take in the `wec` object as the first argument. `wec` is already passed to this function as `self` so I don't know if this is the best solution, but it does make the inputs consistent between the two.

wecopttool/pto.py
jtgrasb: Since we are passing in the whole result directly from the output of `solve`, I changed this to `list[OptimizeResult]` to match that output style.

jtgrasb:
I made updates to this PR and want to summarize the changes here:
- `wec.post_process` and `pto.post_process` now both take in the same inputs (`(wec,res_opt,wave,nsubsteps)`)
  - `wec.post_process` does not use the `wec` input at all, but this was added to be consistent
- Test has been added to test_integration.py to compare result of 2 realizations
- Post-processing functions now always return a list of post-processed results (one for each realization)
  - This is consistent with the `solve` function, which always returns a list of optimized results
- Tutorials have been updated to select the first element in the list of post-processed results when plotting

PR #346, opened 2024-05-07
Add warnings when adding inertia and hydrostatic stiffness automatically

### LINKED ISSUE #345: Warning messages if inertia and hydrostatic stiffness are added automatically
we used to have a function called `linear_hydrodynamics` that you used to add the inertia and stiffness to the BEM output before creating an intrinsic impedance by calling `wot.hydrodynamic_impedance`. This was removed with `capytaine` added in internal calculations of rigid body inertias and hydrostatic stiffnesses and we have added lines in `run_bem` to add these values to the BEM output if they have not already been populated in the `cpy.FloatingBody` object before calling `run_bem`:

https://github.com/sandialabs/WecOptTool/blob/31b03612266e583b9ad7f9db3ab64580274448c5/wecopttool/core.py#L2175-L2180

This is nice in many cases, but if you have a device that is tethered (i.e., the displaced mass of water is not equal to the rigid body mass of the device -- see [`tutorial_2_AquaHarmonics.ipynb`](https://github.com/sandialabs/WecOptTool/blob/main/examples/tutorial_2_AquaHarmonics.ipynb)), then this is not correct. You can still go in and overwrite these values, but it'd probably be nice to add some warning messages in so that people know what's happening.

**Action:** add warning messages in [`run_bem`](https://github.com/sandialabs/WecOptTool/blob/31b03612266e583b9ad7f9db3ab64580274448c5/wecopttool/core.py#L2175-L2180) to note if/when it is auto-populating the rigid body inertia and/or hydrostatic stiffness, e.g.,

```python
if not hasattr(wec_im, 'inertia_matrix'):
    _log.warning('FloatingBody has no inertia_matrix field, using neutral buoyancy assumption to auto-populate')
    wec_im.inertia_matrix = wec_im.compute_rigid_body_inertia(rho=rho) 
```

jtgrasb:
## Description
This PR resolves #345 by adding warning messages when we call `run_bem` and add the inertia matrix and hydrostatic stiffness automatically. Capytaine uses the mass of the FloatingBody if defined. If not, itassumes neutral buoyancy for calculating the [inertia matrix and hydrostatic stiffness](https://capytaine.github.io/stable/user_manual/hydrostatics.html).

### DIFF
--- a/wecopttool/core.py
+++ b/wecopttool/core.py
@@ -2182,11 +2182,9 @@ def run_bem(
                      'be used to auto-populate.')
         wec_im.inertia_matrix = wec_im.compute_rigid_body_inertia(rho=rho)
     if not hasattr(wec_im, 'hydrostatic_stiffness'):
-        _log.warning('FloatingBody has no hydrostatic_stiffness field. ' + 
-                     'If the FloatingBody mass is defined, it will be ' + 
-                     'used for calculating the hydrostatic stiffness here. ' + 
-                     'Otherwise, the neutral buoyancy assumption will ' + 
-                     'be used to auto-populate.')
+        _log.warning('FloatingBody has no hydrostatic_stiffness field. ' +
+                     'Capytaine will auto-populate the hydrostatic ' +
+                     'stiffness based on the provided mesh.'
         wec_im.hydrostatic_stiffness = wec_im.compute_hydrostatic_stiffness(rho=rho, g=g)
     bem_data = solver.fill_dataset(
         test_matrix, wec_im, n_jobs=njobs, **write_info)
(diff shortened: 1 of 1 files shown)


wecopttool/core.py
```diff
@@ -2175,8 +2175,18 @@ def run_bem(
...
     if not hasattr(wec_im, 'inertia_matrix'):
+        _log.warning('FloatingBody has no inertia_matrix field. ' + 
+                     'If the FloatingBody mass is defined, it will be ' + 
+                     'used for calculating the inertia matrix here. ' + 
+                     'Otherwise, the neutral buoyancy assumption will ' + 
+                     'be used to auto-populate.')
         wec_im.inertia_matrix = wec_im.compute_rigid_body_inertia(rho=rho)
     if not hasattr(wec_im, 'hydrostatic_stiffness'):
+        _log.warning('FloatingBody has no hydrostatic_stiffness field. ' + 
+                     'If the FloatingBody mass is defined, it will be ' +
```
ryancoe, line 2186: How would mass be used to calculate the hydrostatic stiffness?

jtgrasb: Fixed. Though the mass can impact the hydrostatic stiffness if the center of rotation is different than the center of mass.

cmichelenstrofer (approved):

PR #362, opened 2024-07-12
Pioneer tutorial updates (empirical impedance model and nonlinear spring)

jtgrasb:
## Description
This PR updates the pioneer tutorial according to the recent work done. The main updates are the use of the `wot.WEC.from_impedance()` method to create the WEC object using an empirical impedance and the addition of the nonlinear spring.
A few notes:
- I left the BEM solver in there to compare to the empirical data.
- The empirical data is loaded from a number of files. It may be worth consolidating these.
- I used the PTO module to create the PTO transfer matrix even though we don't actually use it to define the PTO. Instead we define the generator force, pto power, etc. manually. This could be confusing but left it in for now.
- I made some modifications to the utilities class as well
  - Added option to bode plot to return the axes - this allows us to plot on top of other plots
  - Added option to specify tolerance for sankey diagram. The nonlinear spring case seemed to vary within +- 1% off in terms of the total sum of power flows. The tolerance is by default 0.1% of the optimal excitation, but this case uses 1%.

## Type of PR
- [X] Update to tutorial

## Checklist for PR
- [X] The pull request is from an issue branch (not main) on *your* fork, to the [dev branch in WecOptTool](https://github.com/sandialabs/WecOptTool/tree/dev).
- [ ] All tests pass


michaelcdevin (commented):
Looks good! It's good to see the tutorial updated with the improved model, and the additions to the `utilities` module are a nice touch. I added some comments throughout; they're all minor editorial things.

Regarding the empirical model file imports and consolidation:

* The `Zi_*` and `Znsif_*` files aren't actually used by the empirical model, so we can remove these
* I like the idea of consolidating the empirical data down to a single Python file, `data/pioneer_empirical_data.py`, and have that contain the data variables that we can then just import into this tutorial (e.g., `Zemp_stiffness = 65821.855014021`, `Zemp_imag = np.array([-2094686.18125349, -1046610.28581912, ...])`, etc.). I don't think the file imports are bad though (especially once we remove the unnecessary files), so I'll leave this to your judgment.

I'll approve this, just be sure to address the comments and clean up the data files (include optionally consolidating them) before you merge!

michaelcdevin (approved):

jtgrasb:
Thanks for the review, Michael! I've made the requested updates and merging it now.

PR #363, opened 2024-08-01
Test capytaine lids

jtgrasb:
## Description
This PR uses Capytaine’s new internal lid feature to remove irregular frequency spikes from the BEM results for each of the 4 tutorials. The irregular frequency spikes mostly occur in the added mass and radiation damping but may also impact the excitation forces. The internal lid does not completely remove irregular frequency spikes, but it moves them to higher frequencies (ideally above frequency range of interest). The mesh size and lid position can be tuned to move the spikes outside the frequency range of interest.

## Type of PR
- [X] New feature

## Checklist for PR
- [X] The pull request is from an issue branch (not main) on *your* fork, to the [dev branch in WecOptTool](https://github.com/sandialabs/WecOptTool/tree/dev).
- [ ] All tests pass

## Additional details
WaveBot: The wavebot uses a frequency array containing 10 frequencies up to about 18 rad/s. The impact of irregular frequencies is small, but there is an impact to the result at about 5.6 rad/s where the added mass and radiation damping are lower without the internal lid. In tests of a more discretized frequency array, the irregular frequency spike was found to line up with about 5.6 rad/s, confirming the cause of this discrepancy. Note, I also needed to decrease the mesh size factor from 0.5 to 0.2 to move this frequency spike fully outside the range of interest.
 

Aquaharmonics: The AH tutorial uses a frequency array containing 10 frequencies up to about 8 rad/s. When testing a more discretized frequency array, there is a small spike within the array, but this does not show up in the 10 frequencies as shown below. Still, I applied the lid as it will be useful if users change the frequency array at all.
 

The LUPA example is unique because the float body has a hole in the middle (often referred to as a moonpool). But, the main issue lies in the fact that there is space between the spar and the float (i.e. the hole in the float is larger than the spar). We already have an option to fill the hole in to remove the frequency spike and this works okay. I think a better way to do it is to make the float hole smaller so that its flush with the spar, then apply a lid. I have shown the results for this below and implemented it in the LUPA tutorial.
 

Pioneer: Adding a lid to this tutorial along with reducing the mesh size factor from 0.5 to 0.3 was able to remove the frequency spikes as shown below. This led to a slight change in the power results (~2 W for regular waves).


jtgrasb:
I still want to do some testing of the computation time with the new mesh refinements and the lids, but other than that, the PR is ready to be reviewed.

jtgrasb:
I tested the computation time of the BEM runs for each of the tutorials before and after adding the lids. Both the AquaHarmonics and LUPA tutorials have basically no change in computation time with the lid applied. The WaveBot and Pioneer both have significantly increased computation time, but this is due to the refined mesh (reduced mesh_size_factor from 0.5 to 0.2) to help eliminate frequency spikes.

image

ryancoe:
@dtgaebe will review

examples/tutorial_2_AquaHarmonics.ipynb
```diff
@@ -68,6 +69,7 @@
    "metadata": {},
    "source": [
     "Next we create a `FloatingBody` object from the mesh in Capytaine, which we will use to calculate the hydrodynamics.\n",
+    "Again, we add a lid mesh to remove irregular frequency spikes.\n",
```
dtgaebe, line 72:
```suggestion
    "We add a lid mesh to remove irregular frequency spikes, similar to [Tutorial 1](https://github.com/sandialabs/WecOptTool/blob/main/examples/tutorial_1_WaveBot.ipynb).\n",
```

examples/tutorial_3_LUPA.ipynb
```diff
@@ -39,11 +39,13 @@
     "\n",
     "import gmsh, pygmsh\n",
     "import capytaine as cpy\n",
+    "from capytaine.io.meshio import load_from_meshio\n",
     "import autograd.numpy as np\n",
     "import matplotlib.pyplot as plt\n",
     "import xarray as xr\n",
     "from scipy.optimize import brute\n",
     "\n",
+    "\n",
```
dtgaebe, line 48:
```suggestion
```

dtgaebe (changes requested):
Looks pretty pretty good, maybe some minor changes for someone who just jumps into Tutorial2.

PR #385, opened 2025-01-02
Fix minor friction bug in core.py

### LINKED ISSUE #353: minor bug in `add_linear_friction()`
Pretty sure we're supposed to pass `data`.

https://github.com/sandialabs/WecOptTool/blob/31b03612266e583b9ad7f9db3ab64580274448c5/wecopttool/core.py#L2240-L2241

jtgrasb:
## Description
Resolves #353.


dtgaebe (approved):
should be good now

PR #214, opened 2023-03-30
FOSWEC

rgcoe:
## Description
An example on the FOSWEC (currently bottom fixed)

 - SAND report: https://doi.org/10.2172/1717884
 - Journal paper: https://doi.org/10.1016/j.energy.2021.122485
 - YouTube video: https://youtu.be/OUxbaEC2K6Y

## Checklist for PR
- [x] Authors read the [contribution guidelines](https://github.com/SNL-WaterPower/WecOptTool/blob/main/.github/CONTRIBUTING.md)
- [x] The pull request is from an issue branch (not main) on *your* fork, to the [main branch in WecOptTool](https://github.com/SNL-WaterPower/WecOptTool).
- [x] The authors have given the admins edit access
- [x] All changes adhere to the style guide including PEP8, Docstrings, and Type Hints.
- [ ] Modified the documentation if applicable
- [ ] Modified or added a new test
- [ ] All tests pass

PR #402, opened 2025-04-02
FOSWEC Example

jtgrasb:
## Description
This PR builds off of PR #214 and my analysis of the FOSWEC and comparison to tank test data. Since we already have 4 examples, it may be too much to add a 5th, but there are some unique features that could be included in this example:

- Comparison of BEM and hydrostatic data to experimental data to validate hydrodynamic/impedance model
- Animation of Capytaine degrees of freedom
- Consideration of a multi-dof coupled PTO
- Nonlinear hydrostatics and excitation
- Capacity factor design study

## Type of PR
- New Example


rgcoe:
While I am generally against adding more examples, I agree this is one we should consider - let's discuss as a group.

rgcoe:
Per discussion today, this is definitely something we will aim to include, but want to get MIMO control working properly (see #281, etc.).

PR #406, opened 2025-04-16
Use class for controllers

### LINKED ISSUE #283: Feedback controllers -> class
In reviewing #273, I noticed that we are repeating a fair bit of code and docstrings in the [`pto.py`](https://github.com/sandialabs/WecOptTool/blob/main/wecopttool/pto.py) for the PID controllers. We could potentially use a class to handle this. @cmichelenstrofer pointed out a potential concern that doing this might change the workflow between using an unstructured controller vs. structured controller. Let's discuss this during our upcoming meeting.

rgcoe:
Closes #283


rgcoe:
@cmichelenstrofer - Please take a look and see what you think. This could be considered a breaking change, so keep that in mind.

wecopttool/pto.py
```diff
@@ -55,7 +55,7 @@ class PTO:
     def __init__(self,
         ndof: int,
         kinematics: Union[TStateFunction, ndarray],
-        controller: Optional[TStateFunction] = None,
+        controller: Optional[TStateFunction] = None, #TODO update type
```
cmichelenstrofer, line 58: Needs to be updated. Shouldn't be a `TStateFunction` because it also depends on PTO, right?

cmichelenstrofer (commented):
Examples: 

- Remove `WaveBot_pi.ipynb` (unless you want to make it an official example/tutorial, in which case would need to add Markdown text, etc.
- Update all 4 tutorials to use the new controller class. They run, but now `controller=None` gives a warning. Best practice should be to explicitly specify the controller as the unstructured one and avoid the warning.

wecopttool/controllers.py
```diff
@@ -0,0 +1,233 @@
...
+    
+    @property
+    def ngains(self) -> int:
+        '''Number of controller gains per dof'''
+        return self.proportional + self.integral + self.derivative
+        
+    @property
+    def nstate(self) -> int:
+        '''Total number of controller gains across all DOFs'''
+        return self.ndof * self.ngains
```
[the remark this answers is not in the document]
rgcoe: I don't believe it would handle that case. It'd probably not be too hard to allow the user to pass, e.g., a list of `controller` objects, but I agree this isn't a high priority for us to support in the source.

wecopttool/pto.py
rgcoe: Agreed. So with 4fd895747a1a95f63b5a3414e66057b5fb0342e6, I made a class (`_controller_`) that is a parent to both `pid_controller` and `unstructured_controller`. Do you know how I would say "I want the input argument to be of type `_controller_` and have that be true for the child classes?

cmichelenstrofer: https://stackoverflow.com/a/46092347

```python
from typing import type
controller: Optional[type[_controller_]]
```

PR #416, opened 2025-04-17
tutorial 1: use updated bem object 'hd'

### LINKED ISSUE #412: Use updated bem_data in tutorials
In the tutorials we update `bem_data` and call it `hd`, but then keep using `bem_data`. Is this an error?

rgcoe:
closes #412

### DIFF
--- a/examples/tutorial_1_WaveBot.ipynb
+++ b/examples/tutorial_1_WaveBot.ipynb
@@ -284,11 +284,13 @@
    "metadata": {},
    "source": [
     "Next, we use the utilities class `wecopttool.utilities`, which has functions that help you analyze and design WECs, but are not needed for the core function of WecOptTool.\n",
-    "For example, we calculate the WEC's intrinsic impedance with its hydrodynamic coefficients and inherent inertial properties. We make use of `wecopttool.add_linear_friction()` to convert the `bem_data` into a dataset which contains all zero friction data, because `wecopttool.check_radiation_damping()` and `wecopttool.hydrodynamic_impedance()` excpect a data variable called 'friction'.\n",
+    "For example, we calculate the WEC's intrinsic impedance with its hydrodynamic coefficients and inherent inertial properties. We make use of `wecopttool.add_linear_friction()` to convert the `bem_data` into a dataset which contains all zero friction data, because `wecopttool.check_radiation_damping()` and `wecopttool.hydrodynamic_impedance()` expect a data variable called 'friction'.\n",
     "\n",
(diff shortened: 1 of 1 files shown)


cmichelenstrofer (commented):
@dtgaebe pointed out that the reason it was the way it was:
- `WEC.from_bem()`  does the damping correction for negative or close to zero values
- For plotting the intrinsic impedance we want to have this same correction, so `hd` was corrected. It is the same correction as the one done by the WEC creation, so it is consistent. 

So it was all consistent, and worked well, but it was a bit confusing. The downside of this suggested change is that the user now might think they *have* to correct the BEM radiation as a pre-processing step before creating the WEC object, which is not the case. Thoughts?

If we do want to implement this changes I would rather overwrite `bem_data` instead of calling it a new unintuitive name like `hd`.

rgcoe:
@cmichelenstrofer - What do you think of my proposed version?

cmichelenstrofer:
Looks good! Really like the `bem_data_for_plotting` name, makes it clear. Not sure why the documentaiotn is failing

rgcoe:
Thanks @cmichelenstrofer. It looks like docs are failing due to something with an `npz` file, @jtgrasb is addressing this in #424.

```
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
Cell In[9], line 4
      1 # load in the empirical data
      2 datafile = 'data/pioneer_empirical_data.npz'
----> 4 with np.load(datafile) as empirical_data:
      5     # Access the data within the 'with' block
      6     omega_data = empirical_data['omega_data']
      7     exc_coeff_data = empirical_data['exc_coeff_data']

File /opt/hostedtoolcache/Python/3.12.10/x64/lib/python3.12/site-packages/autograd/tracer.py:48, in primitive.<locals>.f_wrapped(*args, **kwargs)
     46     return new_box(ans, trace, node)
     47 else:
---> 48     return f_raw(*args,**kwargs)

File /opt/hostedtoolcache/Python/3.12.10/x64/lib/python3.12/site-packages/numpy/lib/npyio.py:427, in load(file, mmap_mode, allow_pickle, fix_imports, encoding, max_header_size)
    425     own_fid = False
    426 else:
--> 427     fid = stack.enter_context(open(os_fspath(file),"rb"))
    428     own_fid = True
    430 # Code to distinguish from NumPy binary files and pickles.

FileNotFoundError: [Errno 2] No such file or directory: 'data/pioneer_empirical_data.npz'

You can ignore this error by setting the following in conf.py:

    nbsphinx_allow_errors = True

Info    : [ 90%] Union - Splitting solids                                                                                
Error: Process completed with exit code 1.
```

PR #417, opened 2025-04-17
Inertia matrix fix

### LINKED ISSUE #399: Inertia matrix on immersed part only?
I'm not sure about computing the inertia matrix on the immersed part:

https://github.com/sandialabs/WecOptTool/blob/cde6b775f3ae5248b0b71e13c0e3797cf6a0cf3b/wecopttool/core.py#L2180-L2188

The shape above the free surface should affects the inertia matrix.

jtgrasb:
## Description
Capytaine's `compute_rigid_body_inertia()` function calculates the density as:
``` 
density = rho if self.mass is None else self.mass/volume
```
So, I think we can handle it on the WecOptTool side. The approach would be to define the mass in WecOptTool based on the immersed part, then pass in the full mesh to `compute_rigid_body_inertia()` which will use the mass/volume as the density to calculate the correct inertia values.

Resolves #399 

I've updated the process and warning messages to reflect this.

### DIFF
--- a/wecopttool/core.py
+++ b/wecopttool/core.py
@@ -2182,17 +2182,15 @@ def run_bem(
     if not hasattr(wec_im, 'inertia_matrix'):
         if wec_im.mass is None:
             wec_im.mass = rho*wec_im.copy().keep_immersed_part().volume
+            _log.warning('FloatingBody has no inertia_matrix or mass ' +
+                     'field. The mass will be calculated based on a ' +
+                     'neutral buoyancy assumption. The inertia matrix ' +
+                     'will be calculated assuming a solid and constant ' +
+                     'density body.')
+        else:
             _log.warning('FloatingBody has no inertia_matrix field. ' + 
                      'The FloatingBody mass is defined and will be ' +
-                     'used to determine material density and for ' +
-                     'calculating the inertia matrix.)')
-        else:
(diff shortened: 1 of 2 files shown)


cmichelenstrofer (changes requested):

wecopttool/core.py
```diff
@@ -2177,17 +2177,23 @@ def run_bem(
...
     if not hasattr(wec_im, 'inertia_matrix'):
-        _log.warning('FloatingBody has no inertia_matrix field. ' + 
-                     'If the FloatingBody mass is defined, it will be ' + 
-                     'used for calculating the inertia matrix here. ' + 
-                     'Otherwise, the neutral buoyancy assumption will ' + 
-                     'be used to auto-populate (inertial properties ' +
-                     'based only on the immersed part). We recommend ' +
-                     'inputting the correct inertial properties.')
+        if wec_im.mass is None:
+            wec_im.mass = rho*wec_im.copy().keep_immersed_part().volume
```
mancellin, line 2184: FYI, recent versions of Capytaine have `wec_im.immersed_part()` which is almost exactly `wec_im.copy().keep_immersed_part()` and which has replaced the more dangerous `keep_immersed_part()` in all the examples in the documentation of Capytaine.

PR #431, opened 2025-07-11
Update wave inputs

### LINKED ISSUE #415: BUG: Waves object inconsistency throughout code
**Describe the bug**
The `Waves` xarray now has 3 dimensions: frequencies, directions, and realizations. But all the functions in WecOptTool expect only 2 dimensions. In the WecOptTool solver we itereate through realizations, selecting a subset of the wave xarray at a time, and then all functions work well on these 2-D xarray. Issues:

- The Wave module docstring still says the objects are 2D and does not reflect the addition of the realizations dimension. 
- All functions throughout the code say the format of the waves argument is that from the waves module. That is not correct anymore. They accept a subset (single realization selection). 

I am not sure what the best way to solve this is, without having to edit every single function that uses waves. Just documenting it for now.

jtgrasb:
## Description
This PR resolves #415 by updating the waves object throughout the code to be more consistent. The main changes are:
- Updated all reference to waves object from Dataset to DataArray 
- Updated many functions to use the term `wave` instead of `waves` to indicate when a single realization of the waves object is being used. This is updated for the source and tutorials.
  - The `solve()` and `post_process()` functions still use the full waves object.
- `utilities.calculate_power_flows()` still uses the full waves object but I believe it only calculates the flows for the first realization. We should consider updating this to take in a single realization.

## Type of PR
- [X] Bug fix

## Checklist for PR
- [ ] Authors read the [contribution guidelines](https://github.com/sandialabs/WecOptTool/blob/main/.github/CONTRIBUTING.md)
- [ ] The pull request is from an issue branch (not main) on *your* fork, to the [dev branch in WecOptTool](https://github.com/sandialabs/WecOptTool/tree/dev).
- [ ] The authors have given the admins edit access
- [ ] All changes adhere to the style guide including PEP8, Docstrings, and Type Hints.
- [ ] All tests pass


tests/test_integration.py
```diff
@@ -198,7 +198,7 @@ def test_solve_bounds(bounds_opt, wec_from_bem, regular_wave,
                              obj_fun=p_controller_pto.average_power,
                              nstate_opt=1,
                              x_opt_0=[kplim*0.1],
-                             optim_options={'maxiter': 2e1,
+                             optim_options={'maxiter': 200,
```
cmichelenstrofer, line 201: This PR just updates names, it shouldn't change how the code runs, so not sure why it requires updating the test maxiter. But maybe this is something we should update anyways?

cmichelenstrofer (approved):
Looks good!

tests/test_integration.py
jtgrasb: Yeah, I'm not sure either, but the tests were failing due to the type being "float" rather than "int" so this update was necessary to get the tests to pass.

Issue #324, opened 2024-03-28
Install on an ARM-based mac


ryancoe:
_I have been through this a number of times, and always forget how I solved the problem, so maybe this is useful to someone else._

If you try the following:

```zsh
mamba create -n wot_tmp1
mamba activate wot_tmp1
mamba install -c conda-forge wecopttool
```

you will get

```zsh
Looking for: ['wecopttool']

conda-forge/osx-arm64                                       Using cache
conda-forge/noarch                                          Using cache
pkgs/r/noarch                                                 No change
pkgs/main/osx-arm64                                           No change
pkgs/main/noarch                                              No change
pkgs/r/osx-arm64                                              No change
Could not solve for environment specs
The following packages are incompatible
└─ wecopttool is not installable because there are no viable options
   ├─ wecopttool [2.0.1a1|2.1.0|2.2.0|2.2.2] would require
   │  └─ wavespectra, which does not exist (perhaps a missing channel);
   └─ wecopttool [2.2.2|2.2.3|...|2.6.0] would require
      └─ capytaine, which does not exist (perhaps a missing channel).
```

I think this is because `wavespectra` and `capytaine` both include compiled binaries, and `conda-forge` is not generating those for ARM processors (yet).

ryancoe:
Instead you should do the following:

1. Make sure you have compilers installed on your system (see discussion, e.g., here: https://github.com/capytaine/capytaine/issues/190)
2. Call the following commands (**note this version number for python is based on current version number that `wecopttool` needs per [`pyproject.toml`](https://github.com/sandialabs/WecOptTool/blob/main/pyproject.toml)**)

```zsh
mamba install "python<3.12" pip
pip install wecopttool\[geometry] jupyter
```

ryancoe:
Note that if you want to develop, replace the last two commands with:

```zsh
mamba install "python<3.12" pip pytest ipykernel
pip install -e .\[dev,geometry]
```

cmichelenstrofer:
@ryancoe you cannot install `wavespectra` and `capytaine` from `conda`?

ryancoe:
![Uploading image.png…]()

Issue #35, opened 2022-01-24
Feature request: cache deps. for CI/CD


ryancoe:
https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows

cmichelenstrofer:
Now caching the conda environment with capytaine and wavespectra (and all their dependencies, e.g. numpy, scipy, etc). This will be part of v2.1. When a cache is available this reduces the time to install in ubuntu from ~1.5 minutes to ~15 seconds. On MacOS and Windows there is significant speedup but install still can take long (~5 minutes on Windows!). On Ubuntu almost the netirety of the run time are the tests (~2 minutes).

cmichelenstrofer:
The cache is updated:

- if it is the first time the tests are run that calendar day
- if the `pyproject.toml` file has changed
- if you manually change the `CACHE_NUMBER` value

PR #433, opened 2025-08-11
Autograd -> Jax conversion

jtgrasb:
## Description
Convert autograd to jax. 

Wavebot tutorial is working for the first optimization.


jtgrasb:
To do: 
- Add in functions such as vmap, grad, and jit to increase code performance.
- Resolve MacOS issues on GitHub actions.

jtgrasb:
I added [just-in-time compilation](https://docs.jax.dev/en/latest/jit-compilation.html) to the optimization (objective function, constraints, and relative gradients) using jax.jit which should speed up the code. I also had to change the call to `block_diag()` in the `mimo_transfer_mat()` function to a revised function that is now jittable. There were a couple of assert statements that I commented out for now because assertions related to the dynamic value (some were able to be left in because they were only checking static properties, which is jittable).

Here is the computation time with and without jit for the AquaHarmonics parameter sweep cell:

| Implementation | Time |
| :------- | :------: |
| Autograd | ~ 38 s |
| Jax without jit | ~130 s |
| Jax with jit | ~20 s |

Based on [this issue thread](https://github.com/jax-ml/jax/discussions/5993), scipy does computation on numpy arrays which means the data type keeps getting converted back and forth between jax and numpy arrays. This is why the jax without jit has such a large computation time and the jax with jit computation time is not so much faster than autograd.

jax.vmap:
  - Would need to write functions differently to vectorize. Would be good but a lot of work - save for future work.

To do:
- [x] Resolve MacOS issues on GitHub actions.

rgcoe:
@cmichelenstrofer - I'm trying to debug the CI on this PR for macOS. The key part of [the failing CI log](https://github.com/sandialabs/WecOptTool/actions/runs/17650331026/job/50178598578) seems to be the following, which is either due to some mismatch of `jax` and `jaxlib` (see https://github.com/jax-ml/jax/discussions/14036) and/or something about arm64 (I think this seems likely based on my work below). I'd like your help thinking about this and deciding whether our current workflow of installing partially with `mamba` and partially with `pip` really makes sense.

```
E   AttributeError: partially initialized module 'jax' has no attribute 'version' (most likely due to a circular import)
=========================== short test summary info ============================
ERROR tests/test_core.py - RuntimeError: This version of jaxlib was built using AVX instructions, which your CPU and/or operating system do not support. This error is frequently encountered on macOS when running an x86 Python installation on ARM hardware. In this case, try installing an ARM build of Python. Otherwise, you may be able work around this issue by building jaxlib from source.
ERROR tests/test_integration.py - AttributeError: partially initialized module 'jax' has no attribute 'version' (most likely due to a circular import)
ERROR tests/test_pto.py - AttributeError: partially initialized module 'jax' has no attribute 'version' (most likely due to a circular import)
ERROR tests/test_utilities.py - AttributeError: partially initialized module 'jax' has no attribute 'version' (most likely due to a circular import)
ERROR tests/test_waves.py - AttributeError: partially initialized module 'jax' has no attribute 'version' (most likely due to a circular import)
!!!!!!!!!!!!!!!!!!! Interrupted: 5 errors during collection !!!!!!!!!!!!!!!!!!!!
========================= 5 errors in 99.80s (0:01:39) =========================
```

When I install locally, I do the following trying to replicate the install commands in [pr.yml](https://github.com/sandialabs/WecOptTool/blob/main/.github/workflows/pr.yml).

```
mamba create -n tmp_wot
mamba activate tmp_wot
mamba install python=3.12 capytaine wavespectra
```

This seems to work on the GitHub runner (see, e.g., https://github.com/sandialabs/WecOptTool/actions/runs/17650331026/job/50178598578), but my machine tells me there aren't capytaine and wavespectra binaries for arm64 on conda-forge (see #324 for our previous discussion on this). So I instead do the following, which works fine. 

```bash
mamba create -n tmp_wot
mamba activate tmp_wot
mamba create -n tmp_wot
mamba install pip
cd wecopttool
pip install .
pip install gmsh pygmsh coveralls pytest # we could probably make a special set of optional dependencies in pyproject.toml to avoid this line
coverage run -m pytest
```

Based on all of this, I pushed an update to [pr.yml](https://github.com/sandialabs/WecOptTool/blob/main/.github/workflows/pr.yml) with f1e6e51 to basically do the whole installation via pip. Note that this bypasses the installation of capytaine and wavespectra via mamba which **I believe you added to allow for caching** (see #35)[^1] It seems to fix the problem on macOS, Ubuntu still works fine, and now Windows fails due a Segmentation fault 🤪 -- maybe this is due to these warnings which we have actually been receiving for a while about `cygpath`?

image

[^1]: note that installation via `pip` on macOS and Ubuntu takes only 1-2 minutes (Windows is 8 minutes)

jtgrasb:
For some reason, fully removing mamba and pip installing in editable mode using `-e` worked to get the tests passing on [my own fork](https://github.com/jtgrasb/WecOptTool/pull/8), but didn't work for this PR and the windows test is still failing.

jtgrasb:
After encountering additional GitHub CI errors with uploading to coveralls (errors also [present on dev branch](https://github.com/sandialabs/WecOptTool/actions/runs/18388576420)), I had to edit the workflow further. Here are the changes I made to the pr.yml and push.yml:

- removed caching - this removed a lot of steps that may save time when setting up the environment, but aren't that important because the majority of the time is taken by actually running the tests.
- Replacing `coverage run -m pytest` with `pytest --cov=wecopttool --cov-report=xml` resolved some issues that were occurring when uploading to coveralls. pytest-cov integrates more seamlessly with pytest so this workflow should be better moving forward.

This PR should now be ready for merge.

PR #447, opened 2026-01-22
add 'geometry' dependencies to 'dev'

rgcoe:
The optional dependencies for `dev` should probably include all dependencies to run all tests with no skipping, right?


jtgrasb:
@rgcoe Yes, that's a good point. This should work the same as the prior workflow, since its not actually changing what's installed, just at which step geometry is included. But for installing dev and testing locally, it makes sense to include geometry by default. Merging now.

PR #448, opened 2026-02-06
64 bit floats

rgcoe:
Use 64bit numbers as there may be some cases where this matters


jtgrasb:
This PR specifies the precision (float64) for certain variables which are inputs to the optimization problem, increasing the accuracy of results. May be worth testing in the future to see the impact of this change. Merging now.

PR #451, opened 2026-04-17
Docs build process/systems

rgcoe:
## Description
This is an attempt to simplify and speed up the documentation pages build process.
To accomplish this it 

- Uses sphinx-multiversion to generate documentation pages for `latest` and `dev`
- Add optional arguments for `build_docs.py` to skip running the example notebooks
- Fixes #446 
- Updates/improves instructions for building the docs in `CONTRIBUTING.md`

## Type of PR
- [x] Documentation

PR #452, opened 2026-04-17
Docs improvements

rgcoe:
## Description
Clean up and improvements to documentation pages.

- Add new papers and URLs for old papers to References
- Transition references to Autograd -> Jax

TODO: not sure what this is, but maybe can be removed?  https://github.com/rgcoe/WecOptTool/blob/8577c0ed17596ae47799c97c56077422f246f7ef/wecopttool/core.py#L80-L82


rgcoe:
@dtgaebe - You could add your new animations to this PR if you like, but we (we = you and/or @jtgrasb) should first review/merge #451, as this PR take advantage of and is branched from #451.

dtgaebe:
@rgcoe , @jtgrasb I finally added more description to the theory section with new figure and new solution animation.
I first had to change files to build the docs locally. It should make things more robust for windows. I hope I didn't break anything!

jtgrasb:
@rgcoe @dtgaebe Overall, these changes look good, just a few clarifying points:
- There are sections in the conf.py for both the theory and pseudospectral animations to check if the XXXX_generated variable is true. However, I don't see anywhere that sets the XXXX_generated variables to True or False. Not really an issue, just trying to understand if there is something else this is meant to do such as check if the .pngs are already present?
- `make_pseudospectral_visualizations` is only imported, so the functions are never actually ran and no visualizations generated. This should be updated so the functions are called on import and the visualizations generated.
- Assuming, the above points are fixed to always generate the figures, the png files can be moved to gitignore.

I can make the above change, just documenting to check in with the team before changing and merging PR.

rgcoe:
Thanks very much @jtgrasb; I'd be very grateful if you can carry this over the finish line.

jtgrasb:
Last update I made is to save animations and regenerate them only if missing from folder. That way they don't have to be generated each time documentation is built. 

Merging now.

Issue #453, opened 2026-05-14
Feature request: structured controller in tutorial


rgcoe:
Surprisingly, we don't ever demonstrate how to use structured controllers in tutorials... but we should. Some potentially useful references:

- #283
- #269

PR #457, opened 2026-09-03
Add FOSWEC tutorial

jtgrasb:
## Description
Add FOSWEC tutorial using [Multibody for Everybody](https://github.com/Project-SEA-Stack/Python_Multibody_for_Everybody).
This PR: 
- Adds tutorial 5: FOSWEC notebook, which includes the following important features.
  - Interfaces with Multibody for Everybody (M4E) to setup joints and generate reduced coefficients and matrices.
  - Uses M4E transformation matrix to convert forcing matrices and PTO kinematics to reduced coordinates and convert results back to global coordinates.
  - Verifies integration against WEC-Sim results.
  - Example of static solve - no control variables or objective function, just residual constraint.
  - Example of PTO kinematics matrix for multi-PTO device.
  - Example of a proportional controller. 
    - Satisfies #453, although a saturated controller could still be valuable in the future.
  - Completes a design study on adding top-mounted ballast to replicate work done in the paper.
    - Not exact same as paper to save computation time in tutorial, but results show same trends.
  - Example of multiple solves with different initializations and taking best result to overcome local optima.
- Adds Multibody for Everybody v2 to the project dependencies.
- Adds utilities functions for interfacing with M4E in WecOptTool. Also adds tests for these new functions.

## Type of PR
- New tutorial

## Checklist for PR
- [x] Authors read the [contribution guidelines](https://github.com/sandialabs/WecOptTool/blob/main/.github/CONTRIBUTING.md)
- [x] The pull request is from an issue branch (not main) on *your* fork, to the [dev branch in WecOptTool](https://github.com/sandialabs/WecOptTool/tree/dev).
- [ ] All tests pass

## Additional details
N/A


jtgrasb:
Thank you for the review @riasatmorshed! Summarized here:

Items reviewed and cross-checked
1.	I first derived the FOSWEC multibody configuration in terms of the reduced coordinates to see whether I could manually reproduce the holonomic constraints and the resulting R0 matrix. Everything checks out.
2.	I also checked the geometry-creation subroutine against my own hand-sketched diagram, and that checks out as well.
3.	The overall description and level of detail in the tutorial seem sufficient for a new user to follow.
4.	I also reviewed the construction of the PTO kinematics and drivetrain matrix.

Minor change:
1.	In the plotting cell in Section 2.2, Code-to-code comparison, I suggest adding axs[0].legend() so that the platform curves are identified in the plot.

Potential major issues
1.	In Section 3.1: Top-mounted ballast impact on hydrostatic stiffness, after calculating the modified flap_cg, shouldn’t the mass used subsequently include both the original flap mass and the ballast mass? Additionally, when calculating flap_inertia_hinge, should the mass term be flap_mass + ballast rather than only flap_mass?

2.	I noticed a similar issue in the code cell immediately after Section 3.2.4: Constraints, when     make_floating_body() is called, the mass argument is passed as only flap_mass, which does not appear to include the newly added ballast mass. Should this instead include flap_mass + ballast?

      - Similarly, after adding the ballast mass in the same code-cell, the m0 and J0 should be m0 = [platform_mass, flap_mass+top_mounted_ballast, flap_mass+top_mounted_ballast]

3.	The f_add dictionary does not appear to include f_pto.

jtgrasb:
To address the above review, here are the changes I made:

Minor change:

1. Added legend to axs[0] and made text slightly smaller to avoid covering data.

Bug fixes:

1. Updated mass and inertia to include flap and ballast mass for the hydrostatic stiffness assessment.
2. Updated in the top-mounted ballast loop: mass includes flap and ballast mass. Also updated inertia to be calculated considering both flap and ballast mass.
3. m0 and J0 are updated accordingly. 
4. f_add now included pto.force_on_wec

The last thing I changed is to loop through the optimization with multiple initializations and take the best power result to reduce sensitivity to local optima.

